PR #1, opened 2017-12-10
Only cleanup temporary files when "saveOnly" option is not set

gab1one:
Otherwise the files are deleted before we can import them into ilastik for training, which defeats the whole purpose of this option.


chaubold:
Thanks for the fix!

PR #3, opened 2018-02-15
Junit cases for Hdf5 read/write

ASHISRAVINDRAN:
Added files Hdf5DataSetReaderTest.java, chocolate21.jpg,test.h5.


pom.xml
```diff
@@ -110,5 +110,9 @@
 			<artifactId>scifio</artifactId>
 			<scope>test</scope>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
```
chaubold: This dependency is only required for testing. Please add `<scope>test</scope>` as above, because then users of the resulting `JAR` do not need to have `junit` available.

chaubold:
Thanks for the tests!

pom.xml
ASHISRAVINDRAN: Thank you Carsten for pointing it out. I will correct it.

PR #6, opened 2018-04-05
Argb image hdf5 write

ASHISRAVINDRAN:
Hi Adrian,
As discussed here is the pull request for your reference. I am still in pursuit to check if the contents of the image are correctly converted to ARGB/written as HDF5.


src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -39,7 +38,8 @@
 import net.imglib2.type.numeric.integer.UnsignedShortType;
 
 public class Hdf5DataSetWriterFromImgPlus<T extends Type<T>> {
-	private ImgPlus<T> image;
+    private static int NUM_OF_ARGB_CHANNELS = 4;
+    private ImgPlus<T> image;
 	
     private long nFrames;
 	private long nChannels;
```
wolny: looks like something funky happened to the formatting. Could you reformat this file?

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -245,61 +245,98 @@ private void writeRGB()
...
+                if(image.dimensionIndex(Axes.Z) >= 0)
+                    rai.setPosition(z, image.dimensionIndex(Axes.Z));
+
+                int position =0; //Initializing stack position.
+                for(long c = 0; c < NUM_OF_ARGB_CHANNELS; c++) {
+
+                    // Construct 2D array of appropriate data
+
+                    pixels_byte = new byte[(int) (nCols * nRows)];
+                    if (nChannels==NUM_OF_ARGB_CHANNELS-1 && isFirstSlice ) {
```
wolny: Could you please do all this channel dance before the for-loop. It's a really bad practice to change the loop indexing (var `c` in this case) inside the loop and usually leads to subtle errors.
I'd do the following:
```
if(nChannels == NUM_OF_ARGB_CHANNELS-1) {
  log.warn(...);
  alphaChannelPresent = false;
  nChannels = NUM_OF_ARGB_CHANNELS; // reset channel number
}

for(long c=0; c < nChannels; c++) {
  ...
  if(!alphaChannelPresent && c==0) {
    Arrays.fill(pixel_byte, -1);
  } else {
    fillByteSliceARGB(rai, pixel_byte);
  }
}
```
So you'll create the alpha channel in the h5 file yourself if it's not present for some reason (as in your test case example).

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -245,61 +245,98 @@ private void writeRGB()
...
-//					}
-//					catch (Exception e) {
-//						e.printStackTrace();
-//					}
-//				}
-//			}
-//		}
+        RandomAccess<T> rai = image.randomAccess();
+        boolean isFirstSlice = true;
+        byte[] pixels_byte;
```
wolny: please conform with Java naming conventions (see https://google.github.io/styleguide/javaguide.html e.g.), and use camel-case for var names.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -245,61 +245,98 @@ private void writeRGB()
...
+            {
+                if(image.dimensionIndex(Axes.Z) >= 0)
+                    rai.setPosition(z, image.dimensionIndex(Axes.Z));
+
+                int position =0; //Initializing stack position.
+                for(long c = 0; c < NUM_OF_ARGB_CHANNELS; c++) {
+
+                    // Construct 2D array of appropriate data
+
+                    pixels_byte = new byte[(int) (nCols * nRows)];
```
wolny: Could you please use 2D byte array instead of 1D here, and use the `ncsa.hdf.hdf5lib.H5#H5Dwrite(int, int, int, int, int, java.lang.Object)` instead of `ncsa.hdf.hdf5lib.H5#H5Dwrite(int, int, int, int, int, byte[])`? It's going to be more human readable (in the end images are 2D), and besides that's what's used on the HDF5 examples page, see: https://support.hdfgroup.org/HDF5/examples/api-java.html

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -245,61 +245,98 @@ private void writeRGB()
...
+                            if (dataset_id >= 0) {
+
+                                    H5Dwrite(dataset_id, H5T_NATIVE_UINT8, H5S_ALL, H5S_ALL, H5P_DEFAULT, pixels_byte);
+                            }
+                            else {
+                                throw new HDF5Exception("No active dataset for writing first chunk");
+                            }
+                        }
+                        catch (Exception e) {
+                            e.printStackTrace();
```
wolny: Never use `e.printStackTrace()` in the production code. Please log the exception and re-throw/ignore when necessary!

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -245,61 +245,98 @@ private void writeRGB()
...
+
+                        try {
+                            if (dataspace_id >= 0)
+                            {
+                                H5Sclose(dataspace_id);
+                                dataspace_id = -1;
+                            }
+                        }
+                        catch (Exception e) {
+                            e.printStackTrace();
```
wolny: same here and below

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -245,61 +245,98 @@ private void writeRGB()
...
+                    }
+                    else{
+                        try {
+                            if (dataset_id >= 0) {
+                                dataspace_id = H5Dget_space(dataset_id);
+                                long[] start = {t, z, 0, 0, position}; // tzyxc
+                                H5Sselect_hyperslab(dataspace_id, HDF5Constants.H5S_SELECT_SET, start, null, color_iniDims, null);
+                                int memspace = H5Screate_simple(5, color_iniDims, null);
+                                if (dataspace_id >= 0) {
+                                        H5Dwrite(dataset_id, H5T_NATIVE_UINT8, memspace, dataspace_id, H5P_DEFAULT, pixels_byte);
```
wolny: use `H5.H5DWrite` here, see: https://support.hdfgroup.org/ftp/HDF5/current/src/unpacked/java/examples/datasets/H5Ex_D_Chunk.java as an example

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -521,4 +558,17 @@ private void fillByteSlice(RandomAccess<T> rai, byte[] pixels_byte) {
...
     }
+
+    private void fillByteSliceARGB(RandomAccess<T> rai, byte[] pixels_byte) {
+
+        for (long x = 0; x < nCols; x++) {
+            rai.setPosition(x, image.dimensionIndex(Axes.X));
+            for (long y = 0; y < nRows; y++) {
+                rai.setPosition(y, image.dimensionIndex(Axes.Y));
+                T value = rai.get();
+                pixels_byte[(int) (y * nCols + x)] = (byte) ((ARGBType) value).get();
```
wolny: please use `Math.toIntExact(long)` whenever casting longs to ints in order to prevent hard to find overflow issues

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -521,4 +558,17 @@ private void fillByteSlice(RandomAccess<T> rai, byte[] pixels_byte) {
             }
         }
     }
+
+    private void fillByteSliceARGB(RandomAccess<T> rai, byte[] pixels_byte) {
```
wolny: looks like you can change this signature to `fillbyteSliceARGB(RandomAccess<ARGBType>, ...)` and get rid of the casting inside the method

ASHISRAVINDRAN:
Hi Adrian, Review comments are incorporated and pushed.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -39,400 +39,384 @@
...
+
+                if (nChannels == NUM_OF_ARGB_CHANNELS - 1) {
+                    log.warn("Only 3 channel RGB found. Setting ALPHA channel to -1 (transparent).");
+                    isAlphaChannelPresent = false;
+                }
+                int position = 0; //Initializing stack position.
+                for (long c = 0; c < NUM_OF_ARGB_CHANNELS; c++) {
+                    // Construct 2D array of appropriate data
+                    pixelsByte = new byte[rows][cols];
+                    if (!isAlphaChannelPresent && c == 0) {
```
wolny: I'm still quite uneasy about the `position` index being changed all over this function. How about changing the whole `if` statement to something like this:
```
if(!isAlphaChannelPresent) {
  if(c == 0) {
    // create alpha channel by hand
    for(byte[] row: pixelsByte) {
      Arrays.fill(row, (byte) -1);
    }
  } else {
    if(image.dimensionIndex(Axes.CHANNEL) >= 0) {
      // not a first channel, set the position to c-1
     rai.setPosition(c - 1, image.dimensionIndex(Axes.CHANNEL));
    }
    fillByteSliceARGB(rai, pixelsByte);
  }
} else {
  // all good, we can just use our channel index c as a position
  if(image.dimentionIndex(...) >= 0) {
    rai.setPosition(c, ...);
  }
  fillByteSliceARGB(rai, pixelsByte);
}
```
Still not great, but at least everything is in one place and we avoid the additional `position` index dance

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -39,400 +39,384 @@
...
+                        for (byte[] row : pixelsByte) {
+                            Arrays.fill(row, (byte) -1);
+                        }
+                        position--;
+                    } else {
+                        if (image.dimensionIndex(Axes.CHANNEL) >= 0) {
```
wolny: BTW, is this `if` necessary (here and in other places in the file)? I.e. if the condition is `false` then we never set the position of the `rai`. Is invocation of `fillByteSlice...(rai, pixelsByte)` correct  then?

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -39,400 +39,384 @@
...
+
+                                H5.H5Dwrite(dataset_id, H5T_NATIVE_UINT8, H5S_ALL, H5S_ALL, H5P_DEFAULT, pixelsByte);
+                            } else {
+                                throw new HDF5Exception("No active dataset for writing first chunk");
+                            }
+                        } catch (Exception e) {
+                            log.error("H5D write failed", e);
+                            throw e;
+                        }
+                        try {
```
wolny: Could you please get rid of those boilerplate `try-catch` blocks in this file? They make the code much harder to read and are totally unnecessary in such quantities.
As a rule of thumb:
* use a single try-catch within the function
* if the exception is non-recoverable (i.e. programmer using the API cannot do anything about the exception) then just log and rethrow unchecked exceptions, e.g.:
```
try {
...
} catch(HDF5Exception e) {
  log.error("Failed to write ARGB due to an HDF5 error", e);
  throw new RuntimeException(e); // probably non-recoverable, so rethrow unchecked exception
} catch (Exception e) {
  // cover all the other cases
  log.error("Error writing ARGB", e);
  throw new RuntimeException(e);
}
``` 
See https://github.com/tatsuyaoiw/effective-java/blob/master/chapter-9.md e.g. for more info.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -39,400 +39,384 @@
...
+
+        try {
+            dataspace_id = H5.H5Screate_simple(5, colorIniDims, maxdims);
+        } catch (Exception e) {
+            log.error("H5D dataspace creation failed", e);
+            throw e;
+        }
+
+        try {
+            if ((file_id >= 0) && (dataspace_id >= 0))
```
wolny: btw, I'd expect `H5.H5Screate_simple(...)` to throw an exception instead of returning `dataset_id` equal to -1. Java's not C :smile: so please get rid of those `if(dataspace_id >= 0)` and other such conditions. Have a look at h5 docs to double check

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -441,84 +425,86 @@ else if(hdf5DataType == H5T_NATIVE_FLOAT) {
 
                                 int memspace = H5Screate_simple(5, iniDims, null);
                                 if (dataspace_id >= 0) {
-                                    if(hdf5DataType == H5T_NATIVE_UINT8) {
+                                    if (hdf5DataType == H5T_NATIVE_UINT8) {
```
wolny: could you please use the same API in here, i.e. passing 2d arrays instead of 1d? You might as well leave it for now and tackle it in your subsequent refactoring PR, but make sure to keep the API consistent

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
ASHISRAVINDRAN: Changes scheduled in next PR which basically deals with refactoring everything in the *Hdf5DataSetWriterFromImgPlus* class.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
ASHISRAVINDRAN: Changes scheduled in next PR which basically deals with refactoring everything in the *Hdf5DataSetWriterFromImgPlus* class.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
ASHISRAVINDRAN: Changes scheduled in next PR which basically deals with refactoring everything in the *Hdf5DataSetWriterFromImgPlus* class.

src/test/java/org/ilastik/ilastik4ij/RealUnsignedIntConverter.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public RealUnsignedIntConverter(final double min, final double max) {
+        super(min, max);
+    }
+
+    @Override
+    public void convert(final R input, final UnsignedIntType output) {
+        final double a = input.getRealDouble();
+        output.set(Math.min(2147483647, roundPositive(Math.max(0, ((a - min) / scale * 2147483647)))));
```
imagejan: Use `Integer.MAX_VALUE` instead of `2147483647` to make it more readable.

src/test/java/org/ilastik/ilastik4ij/RealUnsignedIntConverter.java
```diff
@@ -0,0 +1,28 @@
...
+
+import net.imglib2.converter.Converter;
+import net.imglib2.display.AbstractLinearRange;
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.type.numeric.integer.UnsignedIntType;
+
+/**
+ * @author Ashis Ravindran
+ * <p>
+ * Unofficial Converter class to convert images to 32 bit UnsignedIntType.
```
imagejan: This converter should live in a common package like `imglib2` or `imglib2-algorithm` (if it doesn't exist already), no? In any case, the scope of this type of conversion is broader than just `ilastik4ij`.

Also, you might be able to use ImageJ Ops to do the conversion (if you're willing to introduce a dependency on `imagej-ops`).

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -35,490 +10,315 @@
...
-	
-	public void write() 
-	{
-		long[] chunk_dims = {1, 
-            min(nLevs, 256),
-            min(nRows, 256), 
-            min(nCols, 256), 
-            1
+
+        this.nRows = image.dimension(image.dimensionIndex(Axes.Y));
```
wolny: Could you use `int` values everywhere for the x-y dimensions? I.e. just leaving
```
this.nRows = Math.toIntExact(image.dimension(image.dimensionIndex(Axes.Y)));
this.nCols = ...
```
Cause now those values are duplicated which is confusing. Same goes for `nFrames`, `nChannels`, `nLevs` (keep them as the same type, and don't use abbreviations like `nLevs`).

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -35,490 +10,315 @@
...
-            for(long y = 0; y < nRows; y++)
-            {
-                rai.setPosition(y, image.dimensionIndex(Axes.Y));
-                T value = rai.get();
-                pixels_int[(int)(y * nCols + x)] = (int)((UnsignedIntType)value).get();
+
+    private void writeIndividualChannels(int hdf5DataType) {
+        if (nLevs < 1) {
+            log.error("got less than 1 z?");
+            return;
```
wolny: That should be an `IllegalArgumentException`

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
```diff
@@ -35,490 +10,315 @@
...
+        for (int x = 0; x < cols; x++) {
             rai.setPosition(x, image.dimensionIndex(Axes.X));
-            for(long y = 0; y < nRows; y++)
-            {
+            for (int y = 0; y < rows; y++) {
                 rai.setPosition(y, image.dimensionIndex(Axes.Y));
                 T value = rai.get();
-                pixels_byte[(int)(y * nCols + x)] = (byte)(((UnsignedByteType)value).get());
+                if (value instanceof UnsignedByteType) {
+                    pixelArray[y][x] = (E) (Byte) (new Integer(((UnsignedByteType) value).get()).byteValue());
```
wolny: Use `Integer.valueOf(..)` instead of `new Integer(...)`. Same goes for other types below. `new Integer(...)`, `new Long(...)` etc. are deprecated

wolny:
Hi @imagejan, since we're on the topic, could you maybe shed some light on one issue I've discovered while testing @ASHISRAVINDRAN's PR.
I've tested the HDF5 export command with big tiff files. I loaded a 20GB tif stack of the shape `(548, 3913, 4635)` into Fiji with Bio-Formats and tried to export to HDF5. Fiji was silently crashing unfortunately, so I've passed JVM options to do a heap dump on OOM errors. Got the heap dump, analyzed it and it looks like Fiji really tries to load the whole 20GB into memory. I would expect the `ImgPlus` to stream the stack slice-by-slice internally, but instead it creates a list of 10003840 instances of `net.imglib2.img.cell.Cell` each containing 1000 shorts (uint16), which is roughly 20GB.
heapdumpanalysis

Since I wasn't able to hit (https://github.com/ilastik/ilastik4ij/blob/master/src/main/java/org/ilastik/ilastik4ij/IlastikExport.java#L60) while doing the remote debugging, it seems that this happens internally in one of the pre-processing steps in Fiji, before the command gets executed. I'm pretty sure there is a simple way of streaming the file slice-by-slice, e.g. I can easily be done with the Bio-Formats API:
```
        IFormatReader reader = new ImageReader();
        reader.setId("/media/Math-Clinic/brainForSegmentationOneSide.tif");

        for (int series = 0; series < reader.getSeriesCount(); series++) {
            reader.setSeries(series);
            for (int image = 0; image < reader.getImageCount(); image++) {
                byte[] bytes = reader.openBytes(image);
                ...
            }
        }
```
Do you know what needs to be done in order to stream the file slice-by-slice using `ImgPlus` interface? Thanks

imagejan:
@wolny did you have the image open as a virtual stack, or as a normal image window? It might be an issue with converting between the legacy ImageJ1 structures and your `Dataset` input.

I'd assume that the work done by @maarzt in https://github.com/imagej/imagej-legacy/pull/182 could help to improve this.

wolny:
@imagejan thanks for a quick response. Yes, the tiff was opened as a virtual stack with Bio-Formats
bio-formats-import

So if I understands you correctly: the current Fiji version is internally converting from legacy `ImagePlus` to the new `ImgPlus` and that conversion is inefficient... I had a brief look at @maarzt PRs and looks like the (VirtualStackAdapter)[https://github.com/imglib/imglib2-ij/pull/12/files#diff-95c036951492c781899d6982cf00a0ff] would do the job. So when those changes are in the problem will most likely disappear, right?

wolny:
@imagejan is there a way to work around this issue in currently? I'm asking cause the problem is that even if the user runs the export with the machine with enough RAM, the internal conversion is really slow and there is no indication to the user that it's happening under the hood (neither in UI nor in the console logs). I took a thread-dump while the conversion was running and it seems that it's also done in a single thread. Here's the thread-dump:
```

"SciJava-2e720162-Thread-1" #32 prio=5 os_prio=0 tid=0x0000000003629000 nid=0x6098 runnable [0x00007fbb6016b000]
   java.lang.Thread.State: RUNNABLE
        at net.imagej.legacy.translate.GrayPixelHarmonizer.updateDataset(GrayPixelHarmonizer.java:124)
        at net.imagej.legacy.translate.GrayDisplayCreator.makeDataset(GrayDisplayCreator.java:120)
        at net.imagej.legacy.translate.AbstractDisplayCreator.getDataset(AbstractDisplayCreator.java:73)
        at net.imagej.legacy.translate.GrayDisplayCreator.grayCase(GrayDisplayCreator.java:160)
        at net.imagej.legacy.translate.GrayDisplayCreator.makeDisplay(GrayDisplayCreator.java:130)
        at net.imagej.legacy.translate.AbstractDisplayCreator.createDisplay(AbstractDisplayCreator.java:64)
        at net.imagej.legacy.translate.DefaultImageTranslator.createDisplay(DefaultImageTranslator.java:101)
        at net.imagej.legacy.translate.DefaultImageTranslator.createDisplay(DefaultImageTranslator.java:85)
        at net.imagej.legacy.LegacyImageMap.registerLegacyImage(LegacyImageMap.java:273)
        at net.imagej.legacy.display.LegacyImageDisplayService.getActiveImageDisplay(LegacyImageDisplayService.java:119)
        at net.imagej.legacy.display.LegacyImageDisplayService.getActiveDataset(LegacyImageDisplayService.java:130)
        at net.imagej.display.process.ActiveDatasetPreprocessor.getValue(ActiveDatasetPreprocessor.java:72)
        at net.imagej.display.process.ActiveDatasetPreprocessor.getValue(ActiveDatasetPreprocessor.java:56)
        at net.imagej.display.process.SingleInputPreprocessor.process(SingleInputPreprocessor.java:85)
        at org.scijava.module.ModuleRunner.preProcess(ModuleRunner.java:105)
        at org.scijava.module.ModuleRunner.run(ModuleRunner.java:157)
        at org.scijava.module.ModuleRunner.call(ModuleRunner.java:127)
        at org.scijava.module.ModuleRunner.call(ModuleRunner.java:66)
        at org.scijava.thread.DefaultThreadService$3.call(DefaultThreadService.java:238)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

imagejan:
Sorry, I fear I'll be not of much help here. Is it required to have the image open as virtual stack? Or could you simply get the file location (either as a `@Parameter File`, or by asking the `VirtualStack` for its backing file(s)) and then use `DatasetService` or `IOService` to get the ImgPlus from disk, to avoid the legacy environment? It might require to explicitly use ImgLib2`s `Volatile*` structures however. I really don't know. Others might be able to advise, maybe you can open a discussion [on the forum](http://forum.imagej.net/)?

imagejan:
Please also see this related forum discussion: [Wrap ImagePlus VirtualStack into imglib2](http://forum.imagej.net/t/wrap-imageplus-virtualstack-into-imglib2/4154?u=imagejan)

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
ASHISRAVINDRAN: Changed all 5 dimensions to **int** data type. The variable *nLevs* is renamed to **nZ**.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
ASHISRAVINDRAN: Code lifted off. This was part of existing *legacy* code. This is not necessary as the variable *nLevs* (now renamed as **nZ**) is set to **1** in the constructor if the z dimension is less than 1. (*image.dimension(image.dimensionIndex(Axes.Z))* returns -1 if there is just one z dimension in the image. This check is done in the constructor and *1* is set instead of *-1*)

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriterFromImgPlus.java
ASHISRAVINDRAN: Changed this as instructed above. I understand now that these *new Integer(...)*, *new Long(...)* etc. are deprecated from java 9.

src/test/java/org/ilastik/ilastik4ij/RealUnsignedIntConverter.java
ASHISRAVINDRAN: Thank you Jan Eglinger @imagejan  for your review. This converter was not initially created for a numerically accurate conversion, instead just as a wrapper so that the code branch at Hdf5DataSetWriterFromImgPlus class can be tested for 32 bit UnsignedInteger type images. Now that I have made detailed look into the correctness of the conversion, I find that *2147483647* is not the correct as its (2^31-1). Correct number indicating the upper limit of unsigned integer type would be **4294967295**, which is (2^32-1). I have made relevant changes in the *RealUnsignedIntConverter* class so that result of the conversion is numerically correct than before!

src/test/java/org/ilastik/ilastik4ij/RealUnsignedIntConverter.java
ASHISRAVINDRAN: Unfortunately, for some reason, a converter doesn't exist for converting to UnsignedIntType in the package **net.imglib2.converter** where all other converter logics exist. (https://github.com/imglib/imglib2/tree/master/src/main/java/net/imglib2/converter) How can I broaden the scope of *RealUnsignedIntConverter* class? You mean by pull request to the above GitHub repository?

wolny:
@imagejan @ctrueden thanks for the useful tips. Turns out that @ctrueden's suggestion, i.e. loading the tiff as `File > Import > Image...` works as expected (there is no in-memory conversion happening). 

Just for completeness, previously I was loading the 20GB tiff file (tzyxc: 1x548x4635x3913x1) with `Plugins > Bio-Formats > Bio-Formats Importer ` and `View stack with: Data Browser; Memory Management: Use virtual stack`. Then `Plugins > ilastik > Export HDF5` which was causing the in-memory conversion of the whole stack via `net.imagej.legacy.display.LegacyImageDisplayService` crashing my Fiji app.

ctrueden:
Very glad to hear the SCIFIO approach works for you, @wolny.

Sorry that we didn't get the ImageJ Legacy improvements merged before the end of the hackathon. @tpietzsch and @maarzt are working on it—see imglib/imglib2-ij#12, which I anticipate will be merged soon. Then we just need imagej/imagej-legacy#182, which I will do my best to get reviewed + merged ASAP after June 11.

wolny:
Thanks for letting us know @ctrueden. We'll keep monitoring those PRs and ideally build them ourselves and give them a go in order to check the interoperability with BioFormats and our plugin.
On another note @ASHISRAVINDRAN discovered some issues with SCIFIO approach while exporting 'big' tiff files via our plugin. This is related to our discussion but outside of the scope of this particular PR, so we'd double-check his finding and discuss it in a separate PR/issue.

src/test/java/org/ilastik/ilastik4ij/RealUnsignedIntConverter.java
imagejan: Yes, if you have a use case for such a converter that doesn't exist yet in ImgLib2, you should file a pull request there, or ask the developers if there's some other method to achieve what you need.

PR #8, opened 2018-05-13
Status bar for HDF export

ASHISRAVINDRAN:
This code change introduces status bar on FIJI while exporting HDF5.


wolny:
Thanks for the comment @imagejan. Indeed the `Hdf5DataSetWriter` is now obsolete, since we've merged @ASHISRAVINDRAN changes.
`StatusService` is indeed a better idea than static calls to `IJ1`. Both issues are addressed in the following PR: https://github.com/ilastik/ilastik4ij/pull/9

PR #13, opened 2019-02-11
Make import/export macro ready

wolny:
Fixes https://github.com/ilastik/ilastik4ij/issues/14

This PR allows headless processing via Fiji macros. E.g.:
```
import ij.*;
import ij.process.*;
import ij.gui.*;
import java.awt.*;
import ij.plugin.*;
import java.io.File;

public class My_Plugin implements PlugIn {
	public void run(String arg) {
		File dir = new File("<DATASET_DIR>");
		String inputDataset="data";
		String outputDataset="exported_data";
		String ilastikProject = "<ILASTIK_PROJECT_PATH>";
		int i = 0;
		for(File file : dir.listFiles((d, name) -> name.endsWith(".h5"))) {
			IJ.run("Import HDF5", String.format("select=%s datasetname=%s axisorder=tzyxc", file.getAbsolutePath(), inputDataset));
			IJ.run(IJ.getImage(), "Run Pixel Classification Prediction", String.format("saveonly=false projectfilename=%s chosenoutputtype=Probabilities", ilastikProject));
			IJ.run(IJ.getImage(), "Export HDF5", String.format("select=%s/output%d.h5 datasetname=%s compressionlevel=0", file.getParent(), ++i, outputDataset));
		}
	}
}
```


src/main/java/org/ilastik/ilastik4ij/IlastikImport.java
```diff
@@ -21,166 +23,85 @@
...
+        String hdf5FilePath = od.getPath();
+
+        try (IHDF5Reader reader = HDF5Factory.openForReading(hdf5FilePath)) {
+            Map<String, HDF5DataSetInformation> datasets = findAvailableDatasets(reader, "/");
+            List<String> choices = datasets.entrySet()
+                    .stream()
+                    .map(e -> Hdf5Utils.dropdownName(e.getKey(), e.getValue()))
+                    .collect(Collectors.toList());
+
+            GenericDialog gd = new GenericDialog("Select dataset name and axis order");
```
imagejan: Why do you have to resort to using `GenericDialog` here? This seems to me like a step backwards regarding a proper separation of concerns. Isn't it possible to allow the configuration entirely with `@Parameter`s?

wolny: Hi @imagejan. Thanks for the comment. I wasn't able to make it work with `@Parameter`s only, cause I wanted to retain the current UX as much as possible.
selecthdf5

I.e.: allowing the user to specify the *input file*, *dataset name* (drop down), *axis ordering* and whether or not to apply LUT on the resulting image.
One could obviously declare those as:
```
@Parameter
private File inputFile;
@Parameter
private String dataset;
@Parameter
private String axisOrder;
@Parameter
private boolean applyLUT;
```
but that would generate a very simplistic UI where the user would have to specify most of the parameters by hand (e.g. no drop-down).
One could probably still make it work with `@Parameter` and some nice UI, but that would require some convoluted logic like:
```
if(IJ.isMacro()) {
  // headless processing
} else {
  // show UI components
}
```
using `GenericDialog` proved to be a much cleaner approach. Same argument goes for the `IlastikExport.java`.
If you think there is a nice way of retaining the current UX, making the command macro-ready and using `@Parameter`s only, please share :)

imagejan: Alright, I see the point. While there should be a way of dynamically populating the input values, using `DynamicCommand`, I remember there are still some issues in the SciJava framework that might prevent your use case.

Some work was done by @ctrueden and @dietzc in https://github.com/scijava/scijava-common/issues/42#issuecomment-332658377 to make this more flexible, but that's unfinished work. Still I think `ilastik4ij` would be worthwhile to discuss as an example use case in one of the hackathons this or next year.

ctrueden: https://github.com/imagej/tutorials/tree/master/maven-projects/dynamic-commands/src/main/java

ctrueden: What are the dropdown choices intended to be for the dataset? Open images? If so, why does it not work to use `@Parameter Dataset dataset`?

I second @imagejan's point: by reverting to `GenericDialog`, you will introduce a host of problems and limitations in comparison to `@Parameter`. All for the sake of a slightly nicer UI? `@Parameter` can do a lot already—it _can_ do dropdowns with either `choices = {...}` or by using an object type that has entries in the `ObjectService`. Your UI is really not so complicated that a `DynamicCommand` would be needed, or the new NWidget framework linked above, is it? It's four parameters, three of which are trivial?

imagejan: The point is that you want to make the `dataset` choice automatically list the dataset names that are actually present in the selected file, right?

wolny: hey @ctrueden. Re the dropdown choices, it's exactly what @imagejan said: it's a list of `dataset`s available in and H5 chosen by the user. The problem is that the choices are not static, they can be populated only when the `inputFile` parameter becomes available. That's why I wasn't able to make it run completely headless using just `@Parameters` and that's why I fell back to `GenericDialog` which worked out of the box. I'm pretty sure there is a way to achieve it which `@Parameter`s only, but just didn't have the time to explore it more. https://github.com/imagej/tutorials/blob/da48ef6ab1820e74cac696caf32f6725e6a49e29/maven-projects/widget-demo/src/main/java/WidgetDemo.java#L32 doesn't seem to provide a solution. Maybe `DynamicCommand` would be a way to go?

ctrueden: You could use `DynamicCommand`, yeah.

Or you could use the `ScriptService` to launch a subscript that displays the dynamic prompt. See comparison below.

<details><summary>With GenericDialog</summary>

```java
@Override
public void run() {
    OpenDialog od = new OpenDialog("Select export path", "", "export.h5");
    String hdf5FilePath = od.getPath();

    GenericDialog gd = new GenericDialog("Export to HDF5");
    gd.addMessage("Axis order of the exported data set: 'TZYXC'");
    gd.addStringField("ExportPath", hdf5FilePath, 100);
    gd.addStringField("DatasetName", "data");
    gd.addStringField("CompressionLevel", "0");

    gd.showDialog();
    if (gd.wasCanceled()) return;

    hdf5FilePath = gd.getNextString();
    String datasetName = gd.getNextString();
    int compressionLevel = Integer.parseInt(gd.getNextString());

    if (!hdf5FilePath.endsWith(".h5")) {
        IJ.error("Error: HDF5 export file must have '.h5' suffix");
    } else {
        saveImage(hdf5FilePath, datasetName, compressionLevel);
    }
}
```
</details>
<details><summary>With ScriptService</summary>

```java
@Override
public void run() {
	File hdf5File = uiService.chooseFile(new File("export.h5"), FileWidget.SAVE_STYLE);
    String hdf5FilePath = hdf5File.getPath();

	String script = "" +
		"#@ String (visibility = MESSAGE, value = \"Axis order of the exported data set: 'TZYXC'\") message\n" +
		"#@ String (value=\"" + hdf5FilePath + "\") ExportPath\n" +
		"#@ String (value=\"data\") DatasetName\n" +
		"#@ Integer (value=1) CompressionLevel\n";
	Module result = scriptService.run("getInputs.groovy", script, true).get();
	Map<String, Object> values = result.getInputs();

	hdf5FilePath = values.get("ExportPath");
	String datasetName = values.get("DatasetName");
	int compressionLevel = values.get("CompressionLevel");

    if (!hdf5FilePath.endsWith(".h5")) {
        IJ.error("Error: HDF5 export file must have '.h5' suffix");
    } else {
        saveImage(hdf5FilePath, datasetName, compressionLevel);
    }
}
```

</details>

<details><summary>diff ij1.java ij2.java</summary>

```diff
diff --git a/ij1.java b/ij2.java
index 7239233..ba87b82 100644
--- a/ij1.java
+++ b/ij2.java
@@ -1,20 +1,19 @@
 @Override
 public void run() {
-    OpenDialog od = new OpenDialog("Select export path", "", "export.h5");
-    String hdf5FilePath = od.getPath();
+    File hdf5File = uiService.chooseFile(new File("export.h5"), FileWidget.SAVE_STYLE);
+    String hdf5FilePath = hdf5File.getPath();

-    GenericDialog gd = new GenericDialog("Export to HDF5");
-    gd.addMessage("Axis order of the exported data set: 'TZYXC'");
-    gd.addStringField("ExportPath", hdf5FilePath, 100);
-    gd.addStringField("DatasetName", "data");
-    gd.addStringField("CompressionLevel", "0");
+    String script = "" +
+        "#@ String (visibility = MESSAGE, value = \"Axis order of the exported data set: 'TZYXC'\") message\n" +
+        "#@ String (value=\"" + hdf5FilePath + "\") ExportPath\n" +
+        "#@ String (value=\"data\") DatasetName\n" +
+        "#@ Integer (value=1) CompressionLevel\n";
+    Module result = scriptService.run("getInputs.groovy", script, true).get();
+    Map<String, Object> values = result.getInputs();

-    gd.showDialog();
-    if (gd.wasCanceled()) return;
-
-    hdf5FilePath = gd.getNextString();
-    String datasetName = gd.getNextString();
-    int compressionLevel = Integer.parseInt(gd.getNextString());
+    hdf5FilePath = values.get("ExportPath");
+    String datasetName = values.get("DatasetName");
+    int compressionLevel = values.get("CompressionLevel");

     if (!hdf5FilePath.endsWith(".h5")) {
         IJ.error("Error: HDF5 export file must have '.h5' suffix");
```

</details>

Not saying the `ScriptService` API approach is the most elegant thing ever here, but it _is_ UI-agnostic, and would let you dynamically generate the dialog.

PR #21, opened 2019-04-04
Use HTTPS in the POM where feasible

ctrueden:


ctrueden:
This is a stale patch, superseded by 13cf7f9698cc10bb9916a8e342cf321cdd875d12.

Issue #22, opened 2019-05-14
image import seems to ignore gui-specified axis ordering


k-dominik:
I've tried to follow the awesome [Tracking MaMut export tutorial video](https://www.youtube.com/watch?v=kjVqWSSdDmU) on youtube and found that I was unable to load the data into Fiji.

Using the same image file (the drosophila one) and specifying the same axis ordering as in the video I get the following output image:

![meh1](https://user-images.githubusercontent.com/24434157/57708106-12c42980-7669-11e9-9463-f65b966da1cd.png)

Here is the console output for the image, with axisorder `txyzc` specified:

```
[INFO] /volume: GROUP
[INFO] /volume/data: DATASET
[INFO] Found dataset '/volume/data' of type 'uint8'
[INFO] Constructing output image of shape (300, 275, 1, 50, 50). Axis order: 'XYCZT'
[INFO] Loading HDF5 dataset took: 82399
[INFO] Done loading HDF5 file!
[INFO] /volume: GROUP
[INFO] /volume/data: DATASET
[INFO] Found dataset '/volume/data' of type 'uint8'
[INFO] Constructing output image of shape (300, 275, 1, 50, 50). Axis order: 'XYCZT'
[INFO] Loading HDF5 dataset took: 77256
[INFO] Done loading HDF5 file!
```

## Plugin Version:

```
plugins/ilastik4ij-1.5.0.jar

Release date:
05 Mar 2019
```

PR #23, opened 2019-05-15
Fix H5 slice indexing during H5 reading

wolny:
Make sure that the slice read from the H5 file is correctly indexed depending on the XY/YX axis order.


wolny:
Fixes https://github.com/ilastik/ilastik4ij/issues/22

PR #32, opened 2020-01-15
IlastikOptions: load parameters explicitly

stelfrich:
The explicit loading of parameters in the accessors is required so that the parameters always reflect the state provided by a PrefService. We will need this for a custom preference pane in the KNIME ilastik integration to work with the plugin.

### DIFF
--- a/src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
+++ b/src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
@@ -30,14 +30,17 @@ public class IlastikOptions extends OptionsPlugin {
     private int maxRamMb = 4096;
 
     public File getExecutableFile() {
+        load();
         return executableFile;
     }
 
     public int getMaxRamMb() {
+        load();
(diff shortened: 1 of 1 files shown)


tischi:
@wolny Maybe there indeed is an issue with this patch. For example from IntelliJ, the `IlastikOptions` do not work for me anymore:

The simple code below:
```
final ImageJ ij = new ImageJ();

final IlastikOptions ilastikOptions = new IlastikOptions();
ilastikOptions.setExecutableFile( new File("") );
final File executableFile = ilastikOptions.getExecutableFile();
```

Throws 

```
Exception in thread "main" java.lang.NullPointerException
	at org.scijava.command.DynamicCommand.getInfo(DynamicCommand.java:74)
	at org.scijava.options.OptionsPlugin.load(OptionsPlugin.java:94)
	at org.ilastik.ilastik4ij.ui.IlastikOptions.getExecutableFile(IlastikOptions.java:33)
```

@stelfrich 
Maybe some SciJava issue?

PR #38, opened 2020-01-29
explicitly state the ordering of the input axes

k-dominik:
This is a quick fix.
We always write out the data to a tmp folder, and always in the same axes ordering. Currently, this is not understood by ilastik (where the default is `tczyx` vs our plugin `tzyxc`). Therefore stating this when running the workflow resolves this issue. @tischer just to get you involved. Cannot add you as a reviewer.


k-dominik:
merging with @tischer's approval :D

wolny:
:raised_hands: awesome, thanks @k-dominik

PR #40, opened 2020-02-19
Replace --segmentation_image by --binary_image

tischi:


tischi:
@wolny Could you have a look?

wolny:
Hey @tischi. This looks good, just a one-liner :) why does it the parameter name has to be replaced though?

tischi:
I was told by a user that it does not work with `--segmentation_image`, maybe double-check with @k-dominik ?

wolny:
Looks like you're right @tischi. Nice spot!

Issue #46, opened 2020-02-29
Pixel Classification: macro does not record values when segmentation output is selected


k-dominik:
Yah, a weird one... When running pixel classification and selecting probabilities as an output the macro recorder correctly records all parameters:

```
run("Run Pixel Classification Prediction", "projectfilename=/xxxxx/2dcells_pc.ilp inputimage=2d_cells_apoptotic_1channel.png pixelclassificationtype=Probabilities");
```

When segmentations are selected it will only record the following:

```
run("Run Pixel Classification Prediction");
```

I would expect to always see all parameters recorded. Also from the code it is extremely non-obvious how this can happen...

wolny:
Hey @k-dominik,

the only difference in the execution flow between selecting 'Probabilities' vs 'Segmentation' as an output is that selecting segmentation will trigger another command which applies a given LUT to the segmentation output, see: https://github.com/ilastik/ilastik4ij/blob/master/src/main/java/org/ilastik/ilastik4ij/ui/DisplayUtils.java#L28. This executes another ImageJ command via `IJ.run` (see: https://github.com/ilastik/ilastik4ij/blob/master/src/main/java/org/ilastik/ilastik4ij/ui/DisplayUtils.java#L14) and somehow causes the macro recording to skip the necessary invocation parameters.
When I commend the `if`-statement, the macro recording works as expected:
```
run("Run Pixel Classification Prediction", "projectfilename=xxx/pixel_class_2d_cells_apoptotic.ilp inputimage=2d_cells_apoptotic.tif pixelclassificationtype=Segmentation");
```
As a quick fix I'd remove the `if`-statement and always left it to the end user to apply the LUT if needed.

Issue #45, opened 2020-02-29
Pixelclassification: image shown twice


k-dominik:
For some reason, when running pixel classification, the output is opened twice.
Once just as a greylevel image and once as
* with discrete colortable applied in the case of segmentation output
* once with red colormap applied in the case of probability output

I would expect to see only a single output image

PR #47, opened 2020-02-29
Fix macro recording in PixelClassification/ObjectClassification/Tracking 

wolny:
Using `DisplayUtils` with 'Segmentation' as an output type triggers another command which applies the LUT on the output. For some reason this prevents the macro to record the parameters of the command correctly. Skipping the LUT command fixes the issue.

WARN: this will impact PixelClassification/ObjectClassification/Tracking, i.e. Glatsbey LUT won't be applied to the output implicitly and it up to the end user to apply the LUT.

Fixes https://github.com/ilastik/ilastik4ij/issues/46
Fixes https://github.com/ilastik/ilastik4ij/issues/45


k-dominik:
Some more explanation for the root-cause (thx to @imagejan):
https://forum.image.sc/t/scriptparameters-as-arguments-in-a-jython-plugin/30082/5?u=k-dominik

PR #48, opened 2020-03-01
update example macro

k-dominik:
chosenoutputtype -> pixelclassificationtype

somehow our docs were outdated :/


k-dominik:
merging this since I'll use it in a course tomorrow and I have checked that it works.

wolny:
nice catch!

Issue #49, opened 2020-03-02
specify internal path explicitly, internal path implicitly expected to be exported_data for automatic workflow runs


k-dominik:
when in the export settings in ilastik, the name of the dataset is changed, the executor will still expect the output in `exported_data`.

The executor should use the `--output_internal_path` command line parameter?

k-dominik:
already done in https://github.com/ilastik/ilastik4ij/pull/51

PR #51, opened 2020-03-05
Refactor: move common command line arguments to base class

### LINKED ISSUE #50: headless mode invocation should use `--readonly` flag
Because why not?!

k-dominik:
Summary:

* moved all common command line arguments to base class
* explicitly specify `--raw_data` flag for Pixel Classifcation
* specify the `--readonly` flag
* specify the internal export path as `exported_data`

fixes #50, #49


k-dominik:
I'll test it once with all workflows before merging!

Issue #54, opened 2020-07-28
os-specific string indicating ilastik location


k-dominik:
right now we show hints for all three os'es in a single line to indicate ilastik location. This is just very bad UI. OS-specific strings should be shown instead, to keep the text label short.

PR #55, opened 2020-07-28
Make label for ilastik executable system specific

m-novikov:
Closes: #54

### DIFF
--- a/src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
+++ b/src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
@@ -1,5 +1,6 @@
 package org.ilastik.ilastik4ij.ui;
 
+import ij.IJ;
 import org.scijava.options.OptionsPlugin;
 import org.scijava.plugin.Parameter;
 import org.scijava.module.MutableModuleItem;
@@ -36,11 +37,11 @@ public class IlastikOptions extends OptionsPlugin {
         final String os = System.getProperty("os.name").toLowerCase();
         String executableLocation = null;
 
-        if (os.contains("win")) {
+        if (IJ.isWindows()) {
             executableLocation = ILASTIK_PATH_WIN;
-        } else if (os.contains("linux")) {
+        } else if (IJ.isLinux()) {
             executableLocation = ILASTIK_PATH_LINUX;
-        } else if (os.contains("mac")) {
+        } else if (IJ.isMacOSX()) {
             executableLocation = ILASTIK_PATH_MACOS;
         }
 
--- a/src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
+++ b/src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
@@ -2,6 +2,7 @@ package org.ilastik.ilastik4ij.ui;
 
 import org.scijava.options.OptionsPlugin;
 import org.scijava.plugin.Parameter;
+import org.scijava.module.MutableModuleItem;
 import org.scijava.plugin.Plugin;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


k-dominik:
this could potentially also address #57 :)

src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
```diff
@@ -17,10 +18,11 @@
  */
 @Plugin(type = OptionsPlugin.class, menuPath = "Plugins>ilastik>Configure ilastik executable location")
 public class IlastikOptions extends OptionsPlugin {
-    @Parameter(label = "Path to ilastik executable, e.g. " +
-            "MacOS: /Applications/ilastik-1.3.3-OSX.app/Contents/MacOS/ilastik, " +
-            "Windows: C:\\\\Program Files\\ilastik\\ilastik.exe, " +
-            "Linux: /opt/ilastik/run_ilastik.sh")
+    private static final String ILASTIK_PATH_WIN = "C:\\\\Program Files\\ilastik\\ilastik.exe";
+    private static final String ILASTIK_PATH_LINUX = "/opt/ilastik/run_ilastik.sh";
+    private static final String ILASTIK_PATH_MACOS = "/Applications/ilastik-1.3.3-OSX.app/Contents/MacOS/ilastik";
```
k-dominik, line 23: My only suggestion here would be to add some versions to the paths in windows and macos, and make clear that those are examples (as in the original code, it said "e.g. ..."

emilmelnikov:
```java
private static final String[][] EXAMPLE_PATHS = {
    {"win", "C:\\\\Program Files\\ilastik\\ilastik.exe"},
    {"linux", "/opt/ilastik/run_ilastik.sh"},
    {"mac", "/Applications/ilastik-1.3.3-OSX.app/Contents/MacOS/ilastik"},
};

// ...

String path = null;
for (String[] e : EXAMPLE_PATHS) {
    if (os.contains(e[0]) {
        path = e[1];
        break;
    }
}

if (path == null) {
    throw new RuntimeException("Unknown os " + os);
}

getInfo().getMutableInput("executableFile", File.class).setLabel("Path to ilastik executable, e.g. " + path);
```

m-novikov: > make clear that those are examples (as in the original code, it said "e.g. ..."

It already adds "e.g." if the system is identified. See [highlight](https://github.com/ilastik/ilastik4ij/pull/55/files#diff-e2906d73755dc662754edb45de0fc994R49)

k-dominik: > > make clear that those are examples (as in the original code, it said "e.g. ..."
> 
> It already adds "e.g." if the system is identified. See [highlight](https://github.com/ilastik/ilastik4ij/pull/55/files#diff-e2906d73755dc662754edb45de0fc994R49)

sorry, missed that :)

m-novikov:
> @emilmelnikov mentioned that using the old imagej1 api to query for OS should probably be avoided.

I would use it because:
1. There is no alternative in imagej2 API
2. There is no deprecation annotation
3. Personally I doubt that it will ever be deprecated

k-dominik:
> * There is no alternative in imagej2 API
> 
> * There is no deprecation annotation
> 
> * Personally I doubt that it will ever be deprecated

compelling argument :) @emilmelnikov ?

emilmelnikov:
I looked up the implementation of `IJ.isOSX()` and it just checks a system property. So we could just replace it with:

```java
bool isMac = System.getProperty("os.name").toLowerCase().startsWith("mac");
```

m-novikov:
> I looked up the implementation of `IJ.isOSX()` and it just checks a system property. So we could just replace it with:
> 
> ```java
> bool isMac = System.getProperty("os.name").toLowerCase().startsWith("mac");
> ```

I mean why would we implement it ourselves if it's already implemented for us? :)

emilmelnikov:
> I mean why would we implement it ourselves if it's already implemented for us? :)

I don't have a strong opinion about this, but I wouldn't using ImageJ 1 for such a trivial thing.

k-dominik:
At least it's going to be easy to replace if they every get rid of `IJ` :)

PR #62, opened 2020-07-28
Be nice to OSX users and complete path to executable

### LINKED ISSUE #57: specifying ilastik executable on OSX should not require going into the .app
Not 100 % sure if this is the case right now, but judging from the help string, we might expect users to navigate into the `.app` folder to specifiy the executable. This should be done automatically. Specifying the app should suffice.

Note, this should be backward compatible with the settings already done.

k-dominik:
we should not require users to navigate inside the `.app`, since we know were to find the ilastik executable from there.

~~There is one thing I don't like about this solution with the callback: It immediately modifies the path, which is not intuitive  to a user (imo).~~

so I abandoned the callback and implemented a getter for the executable file path in the `AbstractExecutor` class. I just don't want to bother users with the internals of our `.app` at all.

fixes #57 

Todos:

- [x] test this on an osx machine :)


src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
```diff
@@ -18,9 +19,11 @@
 @Plugin(type = OptionsPlugin.class, menuPath = "Plugins>ilastik>Configure ilastik executable location")
 public class IlastikOptions extends OptionsPlugin {
     @Parameter(label = "Path to ilastik executable, e.g. " +
-            "MacOS: /Applications/ilastik-1.3.3-OSX.app/Contents/MacOS/ilastik, " +
+            "MacOS: /Applications/ilastik-1.3.3-OSX.app, " +
```
m-novikov, line 22: We shouldn't change this because it supposed to be handled by #54

k-dominik: okay, will undo this change

src/main/java/org/ilastik/ilastik4ij/executors/ObjectClassification.java
```diff
@@ -27,7 +27,10 @@ public ObjectClassification(File executableFilePath, File projectFileName, LogSe
 
     @Override
     protected List<String> buildCommandLine(Map<String, String> tempFiles, PixelPredictionType secondInputType) {
-        List<String> commandLine = new ArrayList<>(this.baseCommand);
+
+        List<String> commandLine = new ArrayList<>();
```
k-dominik, line 31: ah, I guess I should probably add a getter for this in the base-class...

k-dominik: done

src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
```diff
@@ -62,13 +59,44 @@ public AbstractIlastikExecutor(File executableFilePath, File projectFileName, Lo
...
      *       commandLine.add("...");
      */
     protected abstract List<String> buildCommandLine(Map<String, String> tempFiles, PixelPredictionType pixelPredictionType);
 
+    /*
+     * OS-aware getter for executable file path
+     *
+     * currently adds the internal path to the OSX executable from the .app path.
+     */
+    private String getExecutableFilePath() {
```
emilmelnikov, line 73:
```java
Path p = file.toPath().toAbsolutePath().normalize();
// Convert macOS app bundle path to executable path.
if (IJ.isMacOSX() && p.endsWith(".app")) {
    p = p.resolve(Paths.get("Contents", "MacOS", "ilastik"));
}
return p.toString();
```

src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
```diff
@@ -11,11 +12,8 @@
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.lang.reflect.Array;
```
emilmelnikov, line 15: Leftover import?

src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
emilmelnikov: Or maybe even remove this method and do it inline in `getBaseCommand()`

k-dominik: thanks a lot! Really good to know the nicer/more/more portable modern ways of doing path manipulation

k-dominik: > Or maybe even remove this method and do it inline in getBaseCommand()

I sort of like the story telling aspect of having a dedicated method for it. No need to be bothered with the details when reading `getBaseCommand`

k-dominik:
ok bummer. I tested it on OSX and of course there is a problem. The Parameter is of type `File`, depending on the style. For OSX we'd actually want to select a folder. @m-novikov, in your search for a OS-dependent label, did you stumble upon anything we could exploit there? `style` can be passed to the parameter to achieve a file, vs a directory chooser

k-dominik:
okay, it's really working on osx now, finally. Could you have another look @emilmelnikov ?

src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
```diff
@@ -34,21 +35,28 @@
 
 	@Override
 	public void initialize() {
-        final String os = System.getProperty("os.name").toLowerCase();
         String executableLocation = null;
+        String fileChooserStyle = null;
```
m-novikov, line 39: I think it's better to initialize this to `FileWidget.OPEN_STYLE` and override only in `IJ.isMacOSX()` branch. There is no trailing else so in case it's not one of the mentioned systems it will try setting style to `null`.

k-dominik: Hey @m-novikov, I had a go at your suggestion and slightly refactored some code (I added a method to get this `executableFileItem`). Would you say that's in line with your suggestion?

src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
```diff
@@ -62,13 +60,38 @@ public AbstractIlastikExecutor(File executableFilePath, File projectFileName, Lo
...
 
+    /*
+     * OS-aware getter for executable file path
+     *
+     * currently adds the internal path to the OSX executable from the .app path.
+     */
+    private String getExecutableFilePath() {
+        Path p = executableFile.toPath().toAbsolutePath().normalize();
+        // Convert macOS app bundle path to executable path.
+        if (IJ.isMacOSX() && p.toString().endsWith(".app")) {
```
imagejan, line 77: To avoid a direct dependency on the ImageJ1 class `IJ`, you can also use SciJava's `PlatformUtils`:

```java
import org.scijava.util.PlatformUtils
[...]
if (PlatformUtils.isMac() && p.toString().endsWith(".app")) {
```

k-dominik: thx for keeping the eye out for us @imagejan!

PR #65, opened 2020-07-30
Implements Prediction Mask for Pixel Classification

### LINKED ISSUE #7: Allow optional prediction mask for pixel classification
I suggest adding an optional input to the _Run Pixel Classification Prediction_ node, allowing to choose an image column that contains a prediction mask.

See https://github.com/knime-ip/knip-ilastik/issues/17 for the previous issue opened on `knip-ilastik`.

Tomaz-Vieira:
This still doesn't work as a backwards compatible macro,
since using pixel classification as a macro will require the
value for 'useMask'.

I could not find a way to have this extra parameter with a default
value that would be picked up also by the macro usage. Either
the input is marked as resolved and doesn't show up in the UI,
or it is not marked as resolved, but requires the macro usage to specify
the value interactively.

fixes #7

### DIFF
--- a/src/main/java/org/ilastik/ilastik4ij/ui/IlastikPixelClassificationCommand.java
+++ b/src/main/java/org/ilastik/ilastik4ij/ui/IlastikPixelClassificationCommand.java
@@ -7,6 +7,7 @@ import org.ilastik.ilastik4ij.executors.PixelClassification;
 import org.scijava.ItemIO;
 import org.scijava.app.StatusService;
 import org.scijava.command.Command;
+import org.scijava.command.DynamicCommand;
 import org.scijava.log.LogService;
 import org.scijava.options.OptionsService;
 import org.scijava.plugin.Parameter;
@@ -19,7 +20,7 @@ import java.io.IOException;
 import static org.ilastik.ilastik4ij.executors.AbstractIlastikExecutor.PixelPredictionType;
 
 @Plugin(type = Command.class, headless = true, menuPath = "Plugins>ilastik>Run Pixel Classification Prediction")
-public class IlastikPixelClassificationCommand implements Command {
+public class IlastikPixelClassificationCommand extends DynamicCommand {
 
     @Parameter
     public LogService logService;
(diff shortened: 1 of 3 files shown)


src/main/java/org/ilastik/ilastik4ij/ui/IlastikPixelClassificationCommand.java
```diff
@@ -42,6 +43,21 @@
...
+        useMask = false;
+        //resolveInput("useMask"); //this makes the input not be rendered -.-
+    }
+
+    @Parameter(
+        label = "Prediction Mask",
+        required = false,
+        description = "An image with same dimensions as Raw Data, where the black pixels will be masked out of the predictions"
+    )
+    public Dataset predictionMask;
```
imagejan, line 59: Did you try:
```suggestion
    public Dataset predictionMask = null;
```
to see whether that solves the issue with a default value?

k-dominik:
I'm also hitting an error with the ci, anyone know how to resolve it?

```
ERROR]     Non-resolvable parent POM for org.ilastik:ilastik4ij:1.7.4-SNAPSHOT: Could not transfer artifact org.scijava:pom-scijava:pom:26.0.0 from/to scijava.public (https://maven.scijava.org/content/groups/public): Transfer failed for https://maven.scijava.org/content/groups/public/org/scijava/pom-scijava/26.0.0/pom-scijava-26.0.0.pom 500 Server Error and 'parent.relativePath' points at no local POM @ line 5, column 13
```

imagejan:
> I'm also hitting an error with the ci, anyone know how to resolve it?

That's a temporary issue with server hosting maven.scijava.org.
As far as I know, @ctrueden is on it to fix the problem. I'd recommend waiting until tomorrow and then re-starting the travis build.

k-dominik:
> > I'm also hitting an error with the ci, anyone know how to resolve it?
> 
> That's a temporary issue with server hosting maven.scijava.org.
> As far as I know, @ctrueden is on it to fix the problem. I'd recommend waiting until tomorrow and then re-starting the travis build.

phew, thank you @imagejan I already saw myself editing xml files...

Issue #37, opened 2020-01-29
hdf5 import should read axistags


k-dominik:
No guessing required when the data comes from ilastik, as axistags are written out there. Can fallback to the current way of having to enter them if that data is not present.

PR #66, opened 2020-07-30
New dialog for import

m-novikov:
Closes: #37


k-dominik:
first of all cool that this works at all. I could confirm with a random `h5` file that it first inspects  the file and updates the dialog accordingly.

During my test-ride, I found  that it doesn't seem to work with outputs from ilastik; I'll send you an example.

And second observation is the file selection widget - would it be possible to use something more native looking (like the one that is automatically used by scijava `org.scijava.widget.FileWidget`?)

src/main/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProvider.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private static String parseAxisTags(String jsonString) throws HDF5DatasetEntryProvider.InvalidAxisTagsException {
+        try {
+            JSONObject axisObject = new JSONObject(jsonString);
+            JSONArray axesArray = axisObject.getJSONArray("axes");
+            StringBuilder axisTags = new StringBuilder();
+
+            for (int i = 0; i < axesArray.length(); i++) {
+                JSONObject axisEntry = axesArray.getJSONObject(i);
+                String axisTag = axisEntry.getString("key");
```
Tomaz-Vieira, line 103: I guess if you wanted to be extra nitpicky, you could check if axisTag is in `xyztc` and `throw` otherwise. I'm assuming that ImageJ does not allow arbitrary axis names?

src/main/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProvider.java
```diff
@@ -0,0 +1,128 @@
...
+    public HDF5DatasetEntryProvider(String path, LogService logService) {
+        this.path = path;
+        this.logService = logService;
+    }
+
+    public Vector<DatasetEntry> findAvailableDatasets() {
+        return this.findAvailableDatasets("/");
+    }
+
+    private DatasetEntry getDatasetInfo(String internalPath, IHDF5Reader reader) {
```
Tomaz-Vieira, line 30: nitpick: maybe rename to `getDatasetEntry` ?

src/main/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProvider.java
ctrueden: ImageJ2 does allow arbitrary axis names (`net.imagej.axis.Axes.get(String)`). It's an "extensible enum" so it returns a built-in `Axis` object if one named such exists, otherwise it makes new ones on demand and caches them.

k-dominik:
Hey @m-novikov, I couldn't get this to load a file via the gui (dataset, axes were detected just fine, but no matter what button I clicked the loading got cancelled).

m-novikov:
> Hey @m-novikov, I couldn't get this to load a file via the gui (dataset, axes were detected just fine, but no matter what button I clicked the loading got cancelled).

Sorry messed up flag logic, with the last commit it should work

k-dominik:
Hi @m-novikov,

it's really cool how you managed to get everything into a single dialog. Macro recording and playback works beautifully and is _almost_ fully compatible with the old implementation. The only pathological case I could come up is directly recording a macro in the old version, and executing this with your PR:

```
run("Import HDF5", "select=.../sample_data.h5 datasetname=[/test_image: (17, 37, 127, 137, 7) uint8] axisorder=tzyxc");
```
the "old" plugin recorded the dataset name a little bit differently, including the internal path, shape and data type. I am not sure at the moment how many macros have not removed that information, which will only work if your data doesn't change dimensions (or dtype) over different files, which, to be fair, might often be the case. So I'm not sure users would remove that because it just works. Do you  think it would be hard to support this?

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportMacroOptionsParser.java
```diff
@@ -0,0 +1,39 @@
+package org.ilastik.ilastik4ij.ui;
+
+import ij.Macro;
+
+import javax.crypto.Mac;
```
k-dominik, line 5:
```suggestion
```

leftover... something?

src/test/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProviderTest.java
```diff
@@ -0,0 +1,45 @@
...
+        Vector<HDF5DatasetEntryProvider.DatasetEntry> infos = provider.findAvailableDatasets(TEST_H5_RESOURCE);
+        assertEquals(1, infos.size());
+
+        HDF5DatasetEntryProvider.DatasetEntry info = infos.get(0);
+        assertEquals("/exported_data", info.path);
+        assertEquals("/exported_data: (7, 6, 5, 4, 3) uint16", info.verboseName);
+        assertEquals("tzyxc", info.axisTags);
+    }
+
+    public void testFindAvailableDatasetsWithAxes() {
```
k-dominik, line 35: to feel a little safer may I suggest to have at least one of the test files contain more than a single dataset?

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportModel.java
```diff
@@ -0,0 +1,163 @@
...
+    private final HDF5DatasetEntryProvider entryProvider;
+
+    public IlastikImportModel(LogService logService) {
+        this.logService = logService;
+        propertyChangeSupport = new PropertyChangeSupport(this);
+        entryProvider = new HDF5DatasetEntryProvider(logService);
+    }
+
+    public int getDatasetIdx() {
+        return this.datasetIdx;
```
k-dominik, line 34: I'm not sure about the right code style, but it should be consistent. So either remove `this` here, or add `this` to `getPath` and `getAxisTags`. (probably removing `this` here is preferable)

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportModel.java
```diff
@@ -0,0 +1,163 @@
...
+            return;
+        }
+
+        this.isPathValid = true;
+        String oldPath = this.path;
+        this.path = path;
+
+        try {
+            availableDatasets = entryProvider.findAvailableDatasets(path);
+        } catch (Exception e) {
```
k-dominik, line 60: more a question: is it common in java to catch all Exceptions as opposed to specific exceptions one expects?

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportModel.java
```diff
@@ -0,0 +1,163 @@
...
+        if (this.axisTags.equals(axisTags)) {
+            return;
+        }
+        String oldValue = this.axisTags;
+        this.axisTags = axisTags;
+
+        firePropertyChange(PROPERTY_AXIS_TAGS, oldValue, axisTags);
+    }
+
+    public boolean isAxisTagsValid() {
```
k-dominik, line 137: one could also check that at there are no duplicates for the spacial axes (and time)?

k-dominik (commented):
Really cool that it's working so nicely! I'd say it's a definitive usability improvement. I left some comments/questions.

Todo for me: try it headless.


minor nitpick: not all files end in an empty line.

src/main/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProvider.java
```diff
@@ -0,0 +1,130 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.scijava.log.LogService;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Vector;
+import java.util.stream.Collectors;
+
+public class HDF5DatasetEntryProvider {
```
emilmelnikov, line 16: It seems that this class can be replaced with function `List<DatasetEntry> getDatasetEntries(String path, LogService log)`.

src/main/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProvider.java
```diff
@@ -0,0 +1,130 @@
...
+    static class InvalidAxisTagsException extends Exception {}
+    private final LogService logService;
+
+    public HDF5DatasetEntryProvider(LogService logService) {
+        this.logService = logService;
+    }
+
+    public Vector<DatasetEntry> findAvailableDatasets(String path) {
```
emilmelnikov, line 24: Consider replacing deprecated `Vector` with `List`/`ArrayList` from JCF.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportCommand.java
```diff
@@ -1,94 +1,65 @@
...
-            String datasetName = Hdf5Utils.parseDataset(gd.getNextChoice());
-            rank = datasets.get(datasetName).getRank();
-            String axisOrder = gd.getNextString();
-            boolean applyLUT = gd.getNextBoolean();
-            if (isValidAxisOrder(rank, axisOrder)) {
-                loadDataset(hdf5FilePath, datasetName, axisOrder);
-                if (applyLUT) {
-                    DisplayUtils.applyGlasbeyLUT();
-                }
+    private static IlastikImportDialog dialog = null;
```
emilmelnikov, line 32:
```suggestion
    private static IlastikImportDialog dialog;
```

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportCommand.java
```diff
@@ -1,94 +1,65 @@
...
+    
+    public void run() {
+        IlastikImportModel importModel = new IlastikImportModel(logService);
+        IlastikImportMacroOptionsParser.ParseResult options = IlastikImportMacroOptionsParser.parseOptions(Macro.getOptions());
+
+        importModel.setPath(options.path);
+        importModel.setDatasetPath(options.datasetName);
+        importModel.setAxisTags(options.axisOrder);
+
+        if (!importModel.isValid()) {
```
emilmelnikov, line 42: Consider the early exit:
```java
if (importModel.isValid()) {
    return;
}
// ...

File: src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportCommand.java [hunk 12]
```diff
@@ -1,94 +1,65 @@
...
+            Recorder.recordOption("select", importModel.getPath());
+            Recorder.recordOption("datasetname", importModel.getDatasetPath());
+            Recorder.recordOption("axisorder", importModel.getAxisTags());
         }
+
     }
 
-    private <T extends RealType<T> & NativeType<T>> void loadDataset(String hdf5FilePath, String datasetName, String axisOrder) {
+    public <T extends RealType<T> & NativeType<T>> void loadDataset(String hdf5FilePath, String datasetName, String axisOrder) {
         assert hdf5FilePath != null;
```
emilmelnikov, line 63: Consider replacing asserts with exceptions in public methods: assert checks can be stripped away with a JVM option:
```java
Objects.requireNonNull(hdf5FilePath);
// ...
```

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
```diff
@@ -0,0 +1,226 @@
...
+        );
+        layout.linkSize(SwingConstants.VERTICAL, hdf5PathBrowse, hdf5Path, axisTags, datasetName);
+        contentPanel.setPreferredSize(new Dimension(600, contentPanel.getPreferredSize().height));
+    }
+
+    public void setDatasetNames(Vector<String> names) {
+        datasetNameModel.removeAllElements();
+        for (String name : names) {
+            datasetNameModel.addElement(name);
+        }
```
emilmelnikov, line 100:
```suggestion
        names.forEach(datasetNameModel::addElement);
```

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+    public boolean wasCancelled() {
+        return this.cancelled;
```
emilmelnikov, line 103: Not a big issue, but `this` is usually used only to disambiguate between a method parameter and a class field (mostly in constructors and setters).

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
```diff
@@ -0,0 +1,226 @@
...
+        });
+
+        intializeComponentsLayout();
+        setResizable(true);
+        pack();
+    }
+
+    private boolean isValidAxisOrder(int rank, String axisTags) {
+        if (axisTags.length() != rank) {
+            IJ.error(String.format("Incorrect axis order '%s' for dataset of rank %s", axisTags, rank));
```
emilmelnikov, line 186: `IJ.error` seems out of place in a predicate-like method. Maybe pull this line up and simplify to `return axisTags.length() == rank;`?

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
```diff
@@ -0,0 +1,226 @@
...
+        if (valid) {
+            return VALID_BORDER;
+        } else {
+            return INVALID_BORDER;
+        }
+    }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent evt) {
+        if (evt.getPropertyName().equals(IlastikImportModel.PROPERTY_PATH)) {
```
emilmelnikov, line 202: Java 7 allows `switch` on strings.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
```diff
@@ -0,0 +1,226 @@
...
+            String newPath = (String) evt.getNewValue();
+            if (!this.hdf5Path.getText().equals(newPath)) {
```
emilmelnikov, line 204: What is the reason for the `equals` check? The end result should be the same we `setText(newPath)` unconditionally.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportModel.java
```diff
@@ -0,0 +1,163 @@
...
+            availableDatasets = new Vector<>();
+            this.isPathValid = false;
+            this.setDatasetIdx(-1);
+        }
+
+        firePropertyChange(PROPERTY_PATH, oldPath, path);
+    }
+
+    public void setDatasetPath(String path) {
+        if (!this.isPathValid || availableDatasets.size() == 0) {
```
emilmelnikov, line 70:
```suggestion
        if (!this.isPathValid || availableDatasets.isEmpty()) {
```

src/test/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProviderTest.java
```diff
@@ -0,0 +1,45 @@
...
+    private Context context;
+    private LogService logService;
+
+    private static final String TEST_H5_RESOURCE = "src/test/resources/test.h5";
+    private static final String TEST_AXES_H5_RESOURCE = "src/test/resources/test_axes.h5";
+
+    @Override
+    public void setUp() throws Exception {
+        super.setUp();
+        context = new ImageJ().getContext();
```
emilmelnikov, line 20: I am not sure, but it's not necessary to create `ImageJ`. You can just create a context via constructor.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportModel.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public Vector<String> getAvailableDatasetNames() {
+        Vector<String> result = new Vector<>();
+
+        for (DatasetEntry entry : this.availableDatasets) {
+            result.add(entry.verboseName);
+        }
+
+        return result;
```
emilmelnikov, line 120:
```suggestion
        return availableDatasets.stream().map(e -> e.verboseName).collect(Collectors.toList());
```

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
```diff
@@ -0,0 +1,226 @@
...
+            model.setAxisTags(axisTags);
+
+        } else if (evt.getPropertyName().equals(IlastikImportModel.PROPERTY_AXIS_TAGS)) {
+            String newTags  = (String) evt.getNewValue();
+            if (!this.axisTags.getText().equals(newTags)) {
+                this.axisTags.setText(newTags);
+            }
+        }
+
+        this.hdf5Path.setBorder(getLineBorder(this.model.isPathValid()));
```
emilmelnikov, line 223: Maybe inline the `getLineBorder` method as `setBorder(this.model.isPathValid() ? VALID_BORDER : INVALID_BORDER)` since it's used only in these 2 lines?

src/main/java/org/ilastik/ilastik4ij/hdf5/HDF5DatasetEntryProvider.java
m-novikov: There is a difference between business logic and internal implementation, in this case logService has nothing to do with purpose of this function, so it's better to leave it in the constructor.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportCommand.java
m-novikov: There should not be early access in this case. First, we check if model is valid given options from macro string, if it isn't we show partially filled import dialog.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportDialog.java
m-novikov: seems to be matter of taste in this case

src/main/java/org/ilastik/ilastik4ij/ui/IlastikImportModel.java
m-novikov: Added by this [commit](https://github.com/ilastik/ilastik4ij/pull/66/commits/e25fac4c7c973f690d2e1d17a17cedfe6d91b4a5)

k-dominik:
this looks good to go - @m-novikov, do you want to add anything else here?

PR #69, opened 2020-09-04
Add unified command implementation

emilmelnikov:


emilmelnikov:
Now I don't think this approach is a good idea. It introduces a lot of moving parts, and I am not sure we will be able to support them (or that they will reliably work in the future for that matter).

If we want both user convenience of auto-determining the project type and macros flexibility, we should probably do what @m-novikov did: make a custom "Run Workflow" dialog for ImageJ, and provide commands for each workflow separately (as before).

PR #74, opened 2021-01-11
supply (optional) readonly value explicitly

k-dominik:
the cmd interface of 1.3.2 needs a value with the readonly flag. In order to support this version, we can specify. Works with newer versions, too.

### DIFF
--- a/src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
+++ b/src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
@@ -52,7 +52,7 @@ public abstract class AbstractIlastikExecutor {
             "--output_format=hdf5",
             "--output_axis_order=tzyxc",
             "--input_axes=tzyxc",
-            "--readonly",
+            "--readonly=1",
             "--output_internal_path=exported_data",
             "--input_axes=tzyxc"
         );
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Within ImageJ/Fiji you can install the plugin via the `Help -> Update` menu and 
 
 **Note**: The plugins need Java 1.8, if you see error messages popping up that might be caused by an older Java version.
 
+**Compatibility note:**: We try to keep the plugin backwards compatible. Some versions of ilastik are _not_ compatible with the current plugin: `1.3.3b2`,  `1.3.3rc2`, `1.3.3`, `1.3.3post1`, `1.3.3post2`. _We recommend to update to the latest stable version!. If you need to stick with `1.3.3x` for some reason, please use `1.3.3post3`._
+
 ## User documentation
 
 The ilastik workflow wrappers, as well as importer and exporter, can be found in ImageJ under `Plugins -> ilastik`, 


src/main/java/org/ilastik/ilastik4ij/executors/AbstractIlastikExecutor.java
```diff
@@ -52,7 +52,7 @@ public AbstractIlastikExecutor(File executableFile, File projectFileName, LogSer
             "--output_format=hdf5",
             "--output_axis_order=tzyxc",
             "--input_axes=tzyxc",
-            "--readonly",
+            "--readonly=1",
```
Tomaz-Vieira, line 55: The `--readonly` parameter only started taking a value starting from https://github.com/ilastik/ilastik/commit/a8a3f1985f5c03ff8602cf6dfab965db8a1c1b8f, so this will break with installations older than that.

It is still possible to make this backwards compatible (tough probably awkward; we'd have to get the version number, somehow, hopefully doing something less terrible than grepping for `__version_info__` ). Do you think it's worth the trouble or should be just push the users into using the newest version?

k-dominik: > The --readonly parameter only started taking a value starting from ilastik/ilastik@a8a3f19, so this will break with installations older than that.


Well, that is only part of the story unfortunately :/ If we want to go down the rabbithole...

Historically, ilastik/ilastik@1fd36a383b4a9cc7d9189620c9bc826366d96bef introduced the `--readonly` command line flag that would always have to supply a parameter (2015).
Then it was decided that it defies logic to have a flag pretend to be a parameter, so ilastik/ilastik@bbac5e5c21da19f07a6373d6a8cf7c48317133e4 made it a flag (2019).
Then of course, we arrive at the commit you're referencing, ilastik/ilastik@25cb79f88d3d297410017d007f87807f1035a7fb, preceded by ilastik/ilastik@a8a3f1985f5c03ff8602cf6dfab965db8a1c1b8f with the same basic idea (2020).

In terms of versions, we're now interested in those that treat `--readonly` as a flag.
And of course there are a few beta-versions around that have the flag behaviour, however, the change to an "optional parameter" was also backported to 1.3.3post3. So we have a stable version in 1.3.2, 1.3.3, well and in future 1.4.0 that supports the `--readonly=1`. Some beta versions (and some 1.3.3 stables as well) but there is always a stable version we can point to that even the most cautious users would accept... :)

Tomaz-Vieira: Rats -.-
All right... What if we just executed `ilastik.py --headless` first, captured the usage output, parsed it (terrible, I know, but we'd expect it to be either `[--readonly [READONLY]]` or `[--readonly]`), and then determined if we need the `--readonly` flag or not?

Alternatively, maybe we should tell the user that they should upgrade ilastik and always keep the plugin in sync with the latest stable version of ilastik?

k-dominik: Hmmm - so this would then have to be executed every time we process an image... 

> Alternatively, maybe we should tell the user that they should upgrade ilastik and always keep the plugin in sync with the latest stable version of ilastik?

we could be nice and list which versions of ilastik will now not work anymore (after all, this set is limited) with guides/links to versions that we would recommend them to use if they want to stay at a version close-by?

m-novikov: Maybe we should add version flag for ilastik 1.4.0? So we could query versions in the future and it's first instance where we actually need to distinguish between versions.

Tomaz-Vieira: > Hmmm - so this would then have to be executed every time we process an image...

Maybe only at plugin init time? I'm not sure if the fiji plugin architecture would be friendly to that idea, though

> we could be nice and list which versions of ilastik will now not work anymore

I guess that would be nice too. But if we're not parsing the the error messages or version numbers, then it's possible we'll tell the user to go get a different version even if something unrelated broke

k-dominik: > Maybe we should add version flag for ilastik 1.4.0? So we could query versions in the future and it's first instance where we actually need to distinguish between versions.

sounds like a good plan for the future, but it does not solve the problem at hand. Aesthetically I agree that parsing the output would be the nicest thing to do. But I'm not sure if it is worth the effort (after all it's the first time this issue was reported).

PR #78, opened 2021-04-06
Add multicut

k-dominik:
added executor for pre-trained multicut projects


src/main/java/org/ilastik/ilastik4ij/ui/IlastikMulticutCommand.java
```diff
@@ -0,0 +1,70 @@
...
+
+        try {
+            runClassification();
+        } catch (IOException e) {
+            logService.error("Multicut command failed", e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void runClassification() throws IOException {
```
emilmelnikov, line 65:
```suggestion
    private void runMulticut() throws IOException {
```

src/main/java/org/ilastik/ilastik4ij/ui/IlastikMulticutCommand.java
```diff
@@ -0,0 +1,70 @@
...
+     * Run method that calls ilastik
+     */
+    @Override
+    public void run() {
+
+        if (ilastikOptions == null)
+            ilastikOptions = optionsService.getOptions(IlastikOptions.class);
+
+        try {
+            runClassification();
```
emilmelnikov, line 58:
```suggestion
            runMulticut();
```

src/main/java/org/ilastik/ilastik4ij/executors/ObjectClassification.java
```diff
@@ -29,6 +29,7 @@ public ObjectClassification(File executableFilePath, File projectFileName, LogSe
     protected List<String> buildCommandLine(Map<String, String> tempFiles, PixelPredictionType secondInputType) {
 
         List<String> commandLine = getBaseCommand();
+        commandLine.add("--raw_data=" + tempFiles.get(rawInputTempFile));
```
emilmelnikov, line 32: Leftover?

emilmelnikov (approved):

src/main/java/org/ilastik/ilastik4ij/executors/ObjectClassification.java
k-dominik: hm, nah, I even made it an extra commit. I don't think Object Classification was working properly previously...

emilmelnikov: After looking at the surrounding code, this makes sense.

Maybe also explicitly mention this fix in the merge commit message?

Issue #56, opened 2020-07-28
Better output for object classification


k-dominik:
I've discussed this with @tischi:

Right now the output of object classification is an image with either object probabilities, or an object classification -> pixels are labeled according to the class. In case of touching objects of the same class (that were before extracted with e.g. seeded watershed in the ilastik thresholding applet) it is not possible to tell anymore, that these were separate objects.

A better output would be

 * label-image (object ids) + exported table

this could then be displayed in @tischi's viewer.

PR #84, opened 2021-05-12
Read object features table

emilmelnikov:
Add the ability to read an object features table from object classification workflow.

Also, refactor and simplify all workflows and associated demos.

Closes https://github.com/ilastik/ilastik4ij/issues/56.


wolny (approved):
This is very nicely done, great job @emilmelnikov! The code is much cleaner than what it was before and I'm glad that the workflows can finally be implemented using the `@Plugin` annotations, instead of the oldschool `GenericCommand` (?). 
:+1: :shipit:

k-dominik:
Are these changes backwards compatible (e.g. will old macros still run)? I've tried some of my old ones, but so far without success.

src/main/java/org/ilastik/ilastik4ij/ui/OptionsIlastik.java
```diff
@@ -0,0 +1,56 @@
...
+import org.scijava.options.OptionsPlugin;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+import org.scijava.widget.FileWidget;
+
+import java.io.File;
+
+@Plugin(type = OptionsPlugin.class, menuPath = "Plugins>ilastik>Configure ilastik executable location")
+public class OptionsIlastik extends OptionsPlugin {
+    @Parameter(label = "ilastik executable path", style = FileWidget.OPEN_STYLE)
```
k-dominik, line 13: was previously named `path to ilastik executable`. The rename results in not picking up the "old" value of that variable

k-dominik: well, not sure if it is _this_ rename. in any case, usually it picks up the executable after the update, not this time

src/main/java/org/ilastik/ilastik4ij/workflow/WorkflowCommand.java
```diff
@@ -0,0 +1,121 @@
...
+    public LogService logService;
+
+    @Parameter
+    public StatusService statusService;
+
+    @Parameter
+    public OptionsService optionsService;
+
+    @Parameter(label = "Trained ilastik project file")
+    public File project;
```
k-dominik, line 38:
```suggestion
    public File projectfilename;
```

src/main/java/org/ilastik/ilastik4ij/workflow/WorkflowCommand.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Parameter(label = "Raw input image")
+    public Dataset rawData;
```
k-dominik, line 41:
```suggestion
    public Dataset inputimage;
```

src/main/java/org/ilastik/ilastik4ij/workflow/ObjectClassificationCommand.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+@Plugin(type = Command.class, headless = true, menuPath = "Plugins>ilastik>Run Object Classification Prediction")
+public final class ObjectClassificationCommand<T extends NativeType<T>> extends WorkflowCommand<T> {
+    @Parameter(label = "Pixel probability or segmentation image")
+    public Dataset input2;
```
k-dominik, line 21:
```suggestion
    public Dataset inputProbOrSegImage;
```

emilmelnikov:
> Are these changes backwards compatible (e.g. will old macros still run)? I've tried some of my old ones, but so far without success.
> 
> for Pixel Classification it seems that parameters have been renamed:
> 
> * `projectfilename` -> `project`
> * `inputimage` -> `rawdata`
> 
> I know this sucks, I like the more expressive parameter names, too, but we can't break old macros because of an internal refactor

Good call, I forgot about macros.

src/main/java/org/ilastik/ilastik4ij/ui/OptionsIlastik.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Parameter(min = "-1", label = "maximum number of threads (\"-1\" means no restriction)")
+    private int numThreads = -1;
+
+    @Parameter(min = "256", label = "maximum amount of megabytes")
+    private int maxRamMb = 4096;
+
+    @SuppressWarnings("unused")
+    @Parameter(visibility = ItemVisibility.MESSAGE)
+    private final String examples = "<html>ilastik executable path examples:<dl>" +
```
k-dominik, line 24: We did have this "feature" that only the string relevant for the OS would be shown, any reason you removed that?

k-dominik:
Just to add now that I looked at the code it is really nice to see reduced amount of duplication and boilerplate for the workflows! I tested generating the tables and this works nicely as well. We'll have to add documentation for this though, as one has to configure the csv export in ilastik first - otherwise it won't work. The real power of the table can be unleashed once we allow for exporting the _Object Identities_. That way (I think) one could use feature table as colormaps (somehow) :D. We should maybe ask @tischi how to best do it and provide an example.

emilmelnikov:
> We'll have to add documentation for this though, as one has to configure the csv export in ilastik first - otherwise it won't work.

Would this work?

Screenshot 2021-05-31 at 11 57 45

src/main/java/org/ilastik/ilastik4ij/ui/OptionsIlastik.java
emilmelnikov: Rolled back.

src/main/java/org/ilastik/ilastik4ij/ui/OptionsIlastik.java
emilmelnikov: Wanted to use less magic, but fair point. Now the the correct help string is generated in a static initializer.

emilmelnikov:
> Just to add now that I looked at the code it is really nice to see reduced amount of duplication and boilerplate for the workflows! I tested generating the tables and this works nicely as well. We'll have to add documentation for this though, as one has to configure the csv export in ilastik first - otherwise it won't work. The real power of the table can be unleashed once we allow for exporting the _Object Identities_. That way (I think) one could use feature table as colormaps (somehow) :D. We should maybe ask @tischi how to best do it and provide an example.

I presume we need to change table export in ilastik for this?

k-dominik:
> > The real power of the table can be unleashed once we allow for exporting the _Object Identities_.
> 
> I presume we need to change table export in ilastik for this?

no, we'd need to allow exporting a different source via `--export_source=Object Identities"` command line argument.



> > We'll have to add documentation for this though, as one has to configure the csv export in ilastik first - otherwise it won't work.
> 
> Would this work?
> 
> image

that's perfect :) users don't even need to look into some docs.

emilmelnikov:
> we'd need to allow exporting a different source via `--export_source=Object Identities"` command line argument.

Should we address this in a separate PR?

emilmelnikov:
Superseded by https://github.com/ilastik/ilastik4ij/pull/106.

Object features table could be addressed in a separate PR.

Issue #93, opened 2022-05-05
Request for FIJI plugin to return available datasets for any given HDF5 file


pdd2110:
Hi,

Pretty simple request really. I'm just looking for a command in FIJI that returns the available datasets within a given HDF5 file. I run a lot of the Ilastik IO functions in batch mode, and from time to time I'll get inconsistent (but predictable) datasets names. Attempting to open an unavailable dataset kills the batch run by prompting the user for input. Knowing the available datasets ahead of time would solve this issue.

Looking through the java code on here, I think findAvailableDatasets in the [datasetEntryProvider ](https://github.com/ilastik/ilastik4ij/blob/5d37b8da3be1d9d8d1b55e8454ae6c5e262a19bc/src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetEntryProvider.java#L8) is similar to what I have in mind, but I'm not proficient enough in Java to write my own plugin... Any help would be much appreciated!

P

PR #95, opened 2022-05-06
Add ImageJ command for listing HDF5 datasets

emilmelnikov:
Also refactor DatasetEntry: save shape and dtype as fields, and move verboseName to the toString() method.

Closes https://github.com/ilastik/ilastik4ij/issues/93

### DIFF
--- a/src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetEntryProvider.java
+++ b/src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetEntryProvider.java
@@ -2,7 +2,9 @@ package org.ilastik.ilastik4ij.hdf5;
 
 import hdf.hdf5lib.exceptions.HDF5Exception;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


emilmelnikov:
The new command shows a table with datasets' metadata. However, I'm not sure how to work with such tables from ImageJ macro language.

I tried to use [the macro table API](https://imagej.nih.gov/ij/developer/macro/functions.html) (search for `Table Functions`), but it does not work.

k-dominik:
Heya, this is great, I think the test failures point at some colon missing in new `toString`...:
```
<...dataset_without_axes[:] (64, 64) int64> but was:<...dataset_without_axes[] (64, 64) int64>
```

with regards to the macro table API, maybe someone on the forum can help?

emilmelnikov:
Superseded by https://github.com/ilastik/ilastik4ij/pull/115.

pdd2110:
Hi, thanks for adding this feature!

Any updates on how to access the scijava table via the macro language? I saw that the image.sc forum post hasn't had any activity. Even a `call `function would be helpful.

PR #98, opened 2022-06-14
CI: cache ~/.m2/repository correctly

ctrueden:


k-dominik:
thx @ctrueden

ctrueden:
@k-dominik Weird! My intent was just to make the CI build more efficient, not fix problems. I tried looking back for comparison, to see why the CI build failed before, but the log has already been purged, so I couldn't check.

Issue #101, opened 2023-06-20
Object classification problems with plugin version ilastik4ij-1.8.3-SNAPSHOT.jar


k-dominik:
Currently I cannot run the object classification bit of the plugin (on osx) but get the following error (this is before the dialog to choose images and stuff even shows up):

<details> <summary>traceback</summary>

```
(Fiji Is Just) ImageJ 2.9.0/1.53t; Java 1.8.0_322 [64-bit]; Mac OS X 10.16; 54MB of 12407MB (<1%)
 
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Module threw exception
	at net.imagej.legacy.LegacyService.runLegacyCompatibleCommand(LegacyService.java:308)
	at net.imagej.legacy.DefaultLegacyHooks.interceptRunPlugIn(DefaultLegacyHooks.java:166)
	at ij.IJ.runPlugIn(IJ.java)
	at ij.Executer.runCommand(Executer.java:152)
	at ij.Executer.run(Executer.java:70)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Module threw exception
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at net.imagej.legacy.LegacyService.runLegacyCompatibleCommand(LegacyService.java:304)
	... 5 more
Caused by: java.lang.RuntimeException: Module threw exception
	at org.scijava.module.ModuleRunner.call(ModuleRunner.java:127)
	at org.scijava.module.ModuleRunner.call(ModuleRunner.java:63)
	at org.scijava.thread.DefaultThreadService.lambda$wrap$2(DefaultThreadService.java:225)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	... 1 more
Caused by: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Stack argument out of range: 1
	at net.imglib2.img.VirtualStackAdapter$ImagePlusLoader.get(VirtualStackAdapter.java:194)
[...]

PR #102, opened 2023-06-23
Remove legacy converter and unused dependencies

emilmelnikov:
Accidentally closed https://github.com/ilastik/ilastik4ij/pull/99/, this is the same changeset.

Closes https://github.com/ilastik/ilastik4ij/issues/101.


k-dominik (approved):
Heya, this looks good - awesome that you fixed it. It feels good to be back in a state where we can release whenever.

Also great that you added instructions for the install step.

Let's do a release (and I'm hoping that the "package" step is not needed in the release process somewhere).

PR #106, opened 2023-07-03
Simplify command classes and examples

emilmelnikov:
* Merge commands and executors for less indirection
* Move subprocess running, temporary file management and common parameters to the CommandBase class
* Simplify and merge together demo classes
* Add a simple pixel classification ImageJ macro

Partially supersedes https://github.com/ilastik/ilastik4ij/pull/84 without adding the _features table_ feature.


emilmelnikov:
This PR misses the Tracking demo because data files used in the old demo are not included in the project resources. This demo, along with other workflows, will be added later in another PR.

k-dominik (commented):
Hey @emilmelnikov,

sorry for taking some time to look at this. It looks really nice, I appreciate the cleanup and simplification, as, e.g. with handling of temporary files and the more elegant way of extending/creating new commands!

I still need to do some testing, like, if some old macros that I have still work :)

For future development:
* in order to have a little more uniform flow:
  * add output types to workflows other than Pixel Classification, e.g. `Probabilities`, `Predictions`, `ObjectIDs` to object classification
  * add stage 1 outputs to Autocontext command

src/main/java/org/ilastik/ilastik4ij/cmd/CommandBase.java
```diff
@@ -0,0 +1,203 @@
...
+
+                ProcessBuilder pb = new ProcessBuilder(args);
+                pb.environment().putAll(env);
+
+                logService.info(String.format("%s: starting", logPrefix));
+                logService.info(String.format("%s: arguments = %s", logPrefix, args));
+                logService.info(String.format("%s: environment = %s", logPrefix, env));
+
+                Process proc = pb.start();
+                redirectInputStream(proc.getInputStream(), logService::info);
```
k-dominik, line 123: I guess previously we had `true` there, which marked all outputs as errors :D. It's great that this is now handled properly!! A side effect of this, however, was that the log window would pop up, and it was possible to follow the progress a bit. Is there a way to force-open it? Or do you think it's a bad idea? Would there be another way to indicate that processing is currently going on? Because now it appears as if there's nothing happening until finally the result pops up.

k-dominik:
Okay I ran a few macros and they all still seem to work as before :+1:

src/main/java/org/ilastik/ilastik4ij/cmd/Options.java
```diff
@@ -0,0 +1,41 @@
...
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+import org.scijava.util.PlatformUtils;
+import org.scijava.widget.FileWidget;
+
+import java.io.File;
+
+@Plugin(type = OptionsPlugin.class, menuPath = "Plugins>ilastik>Configure ilastik executable location")
+public final class Options extends OptionsPlugin {
+    @Parameter(label = "Path to ilastik executable", style = FileWidget.OPEN_STYLE)
```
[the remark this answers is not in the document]
emilmelnikov: This should be handled in `initialize()`: `if (PlatformUtils.isMac()) { item.setWidgetStyle(FileWidget.DIRECTORY_STYLE); }`. For me, it works when running `Demo` and selecting _Configure ilastik executable location_ manually.

However, I tried this in Fiji, and, indeed, it doesn't work there, which is unexpected because I used exactly same code from the previous `IlastikOptions` class.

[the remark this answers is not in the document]
k-dominik: Yeah It didn't work before either. Maybe leave it as is, this can be done in a different PR (if at all sensible/possible).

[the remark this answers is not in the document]
emilmelnikov: Yes, this is turns out to be true... I wanted to simplify package structure and remove unnecessary nesting and class name prefixes, but it seems we have to revert to original package names.

Just for the reference, all options are stored via Java [Preferences][prefs] system.

Here is a snippet about OS-dependent locations from ChatGPT:

> On Windows, the preferences are usually stored in the Windows registry under HKEY_CURRENT_USER\Software\JavaSoft\Prefs. Each preference is stored as a key-value pair.

> On macOS, the preferences are stored in the ~/Library/Preferences/com.apple.java.util.prefs.plist file. This file is a property list file that contains the preferences data.

> On Linux and other Unix-like systems, the preferences are stored in a user-specific directory, typically located at ~/.java/.userPrefs. The preferences are stored in XML format within this directory.

On macOS, plist files are binary, but could be converted/viewed as XML as follows: `plutil -convert xml1 ~/Library/Preferences/org.ilastik.ilastik4ij.plist -o - | less`

[prefs]: https://docs.oracle.com/javase/8/docs/api/java/util/prefs/Preferences.html

emilmelnikov:
> Okay I ran a few macros and they all still seem to work as before 👍

Do you know how to read a "non-standard" (e.g. probabilities) image in a macro? For me it did not work because it needed to be imported via BioFormats that shows a pop-up dialog which doesn't work properly in the macro environment.

src/main/java/org/ilastik/ilastik4ij/cmd/CommandBase.java
emilmelnikov: > Would there be another way to indicate that processing is currently going on?

There is `StatusService` that is able to show messages and a progress bar.

Alternatively, we could could try to set a logger level for our messages to `INFO`.

emilmelnikov:
> > Okay I ran a few macros and they all still seem to work as before 👍
> 
> Do you know how to read a "non-standard" (e.g. probabilities) image in a macro? For me it did not work because it needed to be imported via BioFormats that shows a pop-up dialog which doesn't work properly in the macro environment.

After discussing this issue with @k-dominik, it turns out that some of our test TIFF files in this repo are not ImageJ-compatible. Re-exporting them with ImageJ should solve the problem.

emilmelnikov:
Now everything that ilastik prints (both stdout and stderr) goes into the INFO log level. This prevents a log window from popping up when ilastik prints warnings to stderr, which is very common (it occurs if ilastik cannot find any solvers in the system).

src/main/java/org/ilastik/ilastik4ij/cmd/CommandBase.java
emilmelnikov: Log window popup is replaced with an ASCII spinner `|` `/` `-` `\` in the status bar.

README.md
```diff
@@ -11,23 +11,6 @@ The ilastik workflows are invoked by running the ilastik headless mode from the
 Currently, three workflows are wrapped: Pixel classification, Autocontext, Object classification, Boundary-based Segmentation with Multicut, and tracking.
 There is one additional setting showing up in the ImageJ menu, which configures the location of the ilastik binary.
 
-## Contents
```
k-dominik, line 14: Do you mean the "menu" accessible via this button:

image

I didn't know this existed. Do you think users are in general aware of it? (Given users landing at the gihub page looking there might not have used github before)...

src/test/java/org/ilastik/ilastik4ij/Demo.java
```diff
@@ -0,0 +1,81 @@
+package org.ilastik.ilastik4ij;
+
+import net.imagej.Dataset;
+import net.imagej.ImageJ;
+import org.ilastik.ilastik4ij.workflow.ObjectClassificationCommand;
```
k-dominik, line 5: I'm not sure if anyone used this from java before, but we should at least mention in some release notes, that the structure has changed a bit and point to the demo file that you have provided so nicely before releasing :)

btbest (commented):
I've had a browse through the code changes, and this looks like a really good refactor. I haven't actually gone through the setup to try running this now, so let me know if I should as that will probably take me a couple more days...

One leftover question from me regarding the compatibility note in the readme:

> Compatibility note:: We try to keep the plugin backwards compatible. Some versions of ilastik are not compatible with the current plugin: 1.3.3b2, 1.3.3rc2, 1.3.3, 1.3.3post1, 1.3.3post2. We recommend to update to the latest stable version!. If you need to stick with 1.3.3x for some reason, please use 1.3.3post3.

Is that still accurate now; are we able to make statements about compatibility or lack thereof after these changes? As far as I can follow, we are calling ilastik pretty much in the same way as before, so I guess most likely the compatibility hasn't changed. But I also don't really know why the plugin is incompatible with certain 1.3.3 subversions anyway :)

src/main/java/org/ilastik/ilastik4ij/workflow/TrackingCommand.java
```diff
@@ -0,0 +1,36 @@
...
+        return Collections.singletonList("--export_source=Tracking-Result");
+    }
+
+    @Override
+    protected Map<String, Dataset> workflowInputs() {
+        if (ROLE_PROBABILITIES.equals(secondInputType)) {
+            return Collections.singletonMap("prediction_maps", inputProbOrSegImage);
+        }
+        if (ROLE_SEGMENTATION.equals(secondInputType)) {
+            return Collections.singletonMap("binary_image", inputProbOrSegImage);
```
k-dominik, line 32: it might be more useful to have `segmentation_image` here already - I can't imagine that more people are using the tracking with learning workflow, than the "normal" one (or animal tracking).

xref: https://github.com/ilastik/ilastik/issues/2730

src/main/java/org/ilastik/ilastik4ij/workflow/WorkflowCommand.java
```diff
@@ -0,0 +1,282 @@
...
+     * Keys are corresponding command-line arguments without leading dashes.
+     * Values are input datasets.
+     */
+    protected Map<String, Dataset> workflowInputs() {
+        return Collections.emptyMap();
+    }
+
+    @Override
+    public final void run() {
+        logger = logService.subLogger(getClass().getCanonicalName(), LogLevel.INFO);
```
k-dominik, line 90: what is the reason to _not_ use the `logger` declared above, but now use something from `logService`? What is the difference using one vs. the other?

k-dominik (approved):
the spinner is pretty cool :) Especially that it updates the status bar, so even on mouse moves one will be reminded of the fact that ilastik is running!

Let's bump version to at least `1.9.0` when we release this.

README.md
emilmelnikov: On a second thought, I agree that this menu item is not easily discoverable, so we should keep TOC, at least until github makes it's autogenerated menu better.

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5DataSetWriter.java
```diff
@@ -205,11 +205,11 @@ private void createMDArray(IHDF5Writer writer, long[] datasetDims, int[] blockSi
     }
 
     private void writeARGB(IHDF5Writer writer, long[] datasetDims) {
```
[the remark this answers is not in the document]
emilmelnikov: I'm not completely sure, but a cursory look at `writeARGB` tells me that this is because `ARGBType` is a possible element type on an image which encodes 4 channels in an element itself, not in the shape of an image, hence the special handling.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikOptions.java
```diff
@@ -1,85 +1,41 @@
...
-    private static final String ILASTIK_PATH_MACOS = "/Applications/ilastik-1.3.3-OSX.app/Contents/MacOS/ilastik";
-
+public final class IlastikOptions extends OptionsPlugin {
     @Parameter(label = "Path to ilastik executable", style = FileWidget.OPEN_STYLE)
-    private File executableFile = new File("/opt/ilastik/run_ilastik.sh");
+    public File executableFile = new File("/opt/ilastik/run_ilastik.sh");
 
     @Parameter(label = "Number of Threads ilastik is allowed to use.\nNegative numbers means no restriction")
-    private int numThreads = -1;
+    public int numThreads = -1;
```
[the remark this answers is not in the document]
emilmelnikov: Values of all fields annotated with `@Parameter` are somehow populated by the system.

Previously getters just called `load()` before returning field values, but didn't find the reason to do this because `load()` and `save()` are a part of options API, and I think it's better to just read/write them directly. We can't change their names anyway.

src/main/java/org/ilastik/ilastik4ij/workflow/WorkflowCommand.java
emilmelnikov: `LogService#log` logs into the root logger, `Logger#log` logs into some defined namespace (it our case it's fully-qualified class name). Unfortunately, log prefixes are not visible in the _Info_ tab, but the _Log_ tab (or some other log destination) shows them.

PR #110, opened 2023-07-28
Speed up HDF5 processing

emilmelnikov:


src/main/java/org/ilastik/ilastik4ij/ui/IlastikExportCommand.java
```diff
@@ -60,8 +61,7 @@ private <T extends RealType<T> & NativeType<T>> void saveImage(String hdf5FilePa
 
         @SuppressWarnings("unchecked")
         ImgPlus<T> imgPlus = (ImgPlus<T>) input.getImgPlus();
-        new Hdf5DataSetWriter<>(imgPlus, hdf5FilePath, datasetName,
-                compressionLevel, logService, statusService).write();
+        Hdf5.writeDataset(new File(hdf5FilePath), datasetName, imgPlus, compressionLevel);
```
k-dominik, line 64: It would really be nice to get some visual feedback, like the progress bar we had before, when reading/writing the files. (Reading is so fast that it probably doesn't matter, but writing can still take some time).

src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
```diff
@@ -0,0 +1,296 @@
...
+
+        return new ImgPlus<>(
+                ImgView.wrap(view, img.factory()), img.getName(), axes.toArray(new AxisType[0]));
+    }
+
+    /**
+     * Append singleton dimensions that are present in axes but missing from the image.
+     */
+    public static <T extends Type<T>> ImgPlus<T> extendAxes(ImgPlus<T> img, List<AxisType> axes) {
+        RandomAccessibleInterval<T> tmp = img;
```
k-dominik, line 232: for consistency this should also be called `view` as in `permuteAxes`.

k-dominik (commented):
Hey @emilmelnikov,

when I first loaded a file with this code I couldn't believe it! Before it was 2 minutes, and with your changes it feels almost instant. That's really awesome!!

Also thank you for running me through the changes in person, really nice work. I left some comments of what we discovered during the walkthrough.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikExportCommand.java
emilmelnikov: Totally agree. I added callbacks to read and write methods.

However, I'm a bit unsure about the right "large" block size. Currently it's size is selected to be as large as possible while still being able to hold 2^31-1 items, which translates to 2GB–16GB of data, and this obviously results in long delays between callbacks when writing.

What do you think is good value for the "large" block size? Something around 128 MiB?

src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
```diff
@@ -0,0 +1,308 @@
...
+    /**
+     * Guess axes from image dimensions.
+     */
+    public static List<AxisType> guessAxes(long[] dims) {
+        Objects.requireNonNull(dims);
+        // The corresponding default axis ordering logic from ilastik:
+        // https://github.com/ilastik/ilastik/blob/414b6e15a2802ed923ec832776e0f33b1c7d30ae/lazyflow/utility/helpers.py#L76
+        switch (dims.length) {
+            case 2:
+                return new ArrayList<>(Arrays.asList(Axes.X, Axes.Y));
```
btbest, line 189: What's up with creating new ArrayLists here; is the new list from .asList not enough?

src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetDescription.java
```diff
@@ -0,0 +1,116 @@
+package org.ilastik.ilastik4ij.hdf5;
+
+import ch.systemsx.cisd.hdf5.HDF5DataSetInformation;
+import ch.systemsx.cisd.hdf5.IHDF5Reader;
+import hdf.hdf5lib.exceptions.HDF5AttributeException;
+import net.imagej.axis.AxisType;
+import org.ilastik.ilastik4ij.util.ImgUtils;
+import org.json.JSONException;
+
+import java.util.*;
```
btbest, line 10: `.*` makes me uncomfortable

src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetDescription.java
```diff
@@ -0,0 +1,116 @@
...
+                type == that.type &&
+                Arrays.equals(dims, that.dims) &&
+                Objects.equals(axes, that.axes) &&
+                axesGuessed == that.axesGuessed;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(path, type, axes, axesGuessed);
+        result = 31 * result + Arrays.hashCode(dims);
```
btbest, line 102: If this 31 is just an arbitrary seed, I think 42 is obviously a better choice

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5.java
```diff
@@ -0,0 +1,292 @@
...
+    /**
+     * Write image contents to HDF5 dataset, creating/overwriting file and dataset if needed.
+     * <p>
+     * Only 2D-5D datasets with types enumerated in {@link DatasetType} are supported.
+     * As a special case, {@link ARGBType} is supported too, but its use is discouraged.
+     * <p>
+     * if axes are specified, image will be written in the specified axis order.
+     * <p>
+     * <p>
+     * If not null, callback will be invoked each time a block is about to be written.
```
btbest, line 182:
```suggestion
     * If not null, callback will be invoked each time a block is about to be written. This is used to provide feedback via progress bar.
```

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5.java
```diff
@@ -0,0 +1,292 @@
...
+                     compressionLevel)) {
+
+            long gridIndex = 0;
+            while (gridCursor.hasNext()) {
+                callback.accept(gridIndex++, gridSize);
+                RandomAccessibleInterval<T> block = gridCursor.next();
+                Cursor<T> blockCursor = Views.flatIterable(block).cursor();
+                long[] currBlockDims = reversed(block.dimensionsAsLongArray());
+                long[] offset = reversed(block.minAsLongArray());
+                type.writeCursor(blockCursor, writer, dataset, currBlockDims, offset);
```
btbest, line 247: It's really weird that we `read` and `write` from the dataset-type. I get that the point is to keep the type-specific logic all in DatasetType, and I would agree that that has higher priority. But would be nice if there is a way to keep the type-specific logic there but also not have all of this functionality tacked onto an enum.

btbest (approved):

src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
emilmelnikov: `Arrays.asList` returns a fixed-size list that can't be appended to.

src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetDescription.java
emilmelnikov: This is what IntelliJ does by default (at least for me). Do you think we need to forbid wildcard imports?

src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetDescription.java
emilmelnikov: This is an odd prime that also has [good performance properties](https://stackoverflow.com/questions/5154970/how-do-i-create-a-hashcode-in-net-c-for-a-string-that-is-safe-to-store-in-a/5155015#comment51056393_5155015). The same constant is used in `Arrays.hashCode` in OpenJDK.

emilmelnikov: > I hope that `equals/hashCode` was auto-generated by your IDE.

Of course, IntelliJ's _Generate equals() and hashCode()_ is amazing! Although proper value objects are long overdue in Java...

src/main/java/org/ilastik/ilastik4ij/hdf5/Hdf5.java
```diff
@@ -0,0 +1,292 @@
...
+
+            HDF5DataTypeInformation typeInfo = info.getTypeInformation();
+            type = DatasetType.ofHdf5(typeInfo).orElseThrow(() ->
+                    new IllegalArgumentException("Unsupported dataset type " + typeInfo));
+
+            int[] blockDims = info.getStorageLayout() == HDF5StorageLayout.CHUNKED ?
+                    reversed(info.tryGetChunkSizes()) : largeBlockDims(dims);
+
+            try (HDF5DataSet dataset = reader.object().openDataSet(path)) {
+                if (IntStream.range(0, dims.length).allMatch(i -> dims[i] == blockDims[i])) {
```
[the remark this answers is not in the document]
emilmelnikov: I wanted to catch a special case when HDF5 is chunked, but the chunk shape is equal to dataset shape, but this case should be extremely rare, so good catch!

However, I think I'd want to change this a little bit: chunks are usually quite small, and we don't want to create `CellImg` with such small cells, so it's probably better to set block (cell) size to be some multiple of HDF5 chunk size.

src/main/java/org/ilastik/ilastik4ij/ui/IlastikExportCommand.java
```diff
@@ -60,8 +61,7 @@ private <T extends RealType<T> & NativeType<T>> void saveImage(String hdf5FilePa
...
 
         Instant finish = Instant.now();
         long timeElapsed = Duration.between(start, finish).toMillis();
```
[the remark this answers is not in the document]
emilmelnikov: I added timing logs in `WorkflowCommand`. Export and import deserve their own PR anyway, so I think it's better to add timing logs there.

src/main/java/org/ilastik/ilastik4ij/hdf5/DatasetDescription.java
emilmelnikov: I think comment is not really that necessary because it's the (somewhat) default way of creating value classes that override `equals()`, `hashCode()`, and `toString()`. This is described in _Effective Java_, and IntellJ knows this as well.

emilmelnikov:
I've made some more changes.

Now HDF5 chunk size and imglib2 cell size are always separate. When reading chunked HDF5 file, it's chunk size no longer used as a cell size because chunk size is too small. When writing, cell size is chosen to be at most 256 MiB, which should be a good compromise between interactivity and performance. Write cell shape is now always either `XY` or `XYZ` with singleton dimensions for other axes.

When writing, callbacks now trigger status bar updates, so users can see the progress. Callback 
signature is changed: now it accepts `long` which is the total number of bytes written (it is easier to deal with bytes because we can precalculate the total number of bytes we need to read/write in advance, and use this for progress bar updates).

emilmelnikov:
Side note: imglib2 recently added [`PrimitiveBlocks`](https://github.com/imglib/imglib2/pull/330) for copying an arbitrary subregion into a raw array. I've benchmarked it for writing a "simple" dataset, and it is faster by 20–40% depending on the specific axis order! However, if the target dataset is a more complex view, it switches to fallback implementation which, in my tests, resulted in out-of-heap error. Therefore, I left existing implementation for now.

k-dominik (approved):
Really nice! With all the eyes this PR got and all the praise, too, let's not delay merging it (pls don't squash).

The only minor cosmetic issue I noticed is that the progress bar is not reset to zero when writing during workflow processing. It's reset after processing is done though, so, also fine to merge without fixing it.

PR #111, opened 2023-08-11
More workflow refactoring

emilmelnikov:
* Extract subprocess running logic into Subprocess
* TempDir is now a separate class
* Extract Spinner into more general StatusBar
* Merge WorkflowCommand core logic into a single method


k-dominik (approved):
lgtm, thx for in-person walk-through :)

PR #112, opened 2023-08-18
Add demo for converting from/to ImagePlus

emilmelnikov:
ImagePlus is still heavily used by ImageJ1 users.

Also narrowed generic type bounds in Workflow classes to `NativeType<T> & RealType<T>`: it's more precise and allows one to call ImageJFunctions.wrap without casting.


emilmelnikov:
cc @tischi

k-dominik (approved):
lgtm

PR #126, opened 2024-03-21
Added axes serialization to json for export

tischi:


src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
```diff
@@ -328,11 +328,31 @@ public static List<AxisType> parseAxes(String json) {
...
-            jsonAxis.put("key", axis.getLabel());
-            jsonAxesArray.put( jsonAxis );
+            axis.isSpatial();
+            String label = axis.getLabel().toLowerCase();
+            int vigraType;
+            String vigraKey;
+            if (axis.isSpatial()){
+                vigraType = 2;
+                vigraKey = label;
+            } else if (label.equals("time")) {
```
k-dominik, line 343: Hey @tischi, I've added a typeflag that is necessary for vigra to properly recognize the axes. Do you think that one can rely on Fiji/ImageJ always having the strings "TIME" and "CHANNEL" as labels for channel and time axes?

tischi: I would think so, but we would need to ask the SciJava folks to be sure...
Is there some `axis.isTemporal()`?

k-dominik: I only could find `isSpacial` unfortunately [ref](https://javadoc.scijava.org/ImageJ2/net/imagej/axis/AxisType.html).

src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
```diff
@@ -325,6 +325,42 @@ public static List<AxisType> parseAxes(String json) {
...
 
+    public static String axesToJSON(List<AxisType> axes)
+    {
+        JSONArray jsonAxesArray = new JSONArray();
+        List<AxisType> axes_for_serialization = new ArrayList<>(axes);
+        Collections.reverse(axes_for_serialization);
+        for (AxisType axis : axes_for_serialization)
+        {
+            JSONObject jsonAxis = new JSONObject();
+            axis.isSpatial();
```
btbest, line 336: Does this do anything here?

btbest (commented):
Oh boy we've had this neat feature sitting around for ages :facepalm:

I think it looks good - though I didn't bother to actually compile and try it now 😶 

ilastik's own hdf5 image export includes two more keys in the axistags json: `
```
"axes": [
    {
      "key": "y",
      "typeFlags": 2,
      "resolution": 0,
      "description": ""
    },
    //...
]
```
These are `0` and `""` for all axes in the example I exported, so I would assume they are optional.

src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
k-dominik: ha, good find - looks like an oopsie on my side ;)

src/main/java/org/ilastik/ilastik4ij/util/ImgUtils.java
k-dominik: okay, let's conclude -  no isTemporal and we stay full of trust that the label is always called time :)

k-dominik:
Thank you for having a look @btbest 

regarding your question on missing `resolution` and `description` - these are optional ([default parameters](https://ukoethe.github.io/vigra/doc-release/vigranumpy/#vigra.AxisInfo) for `AxisInfo` construction are in place).

<details><summary>test script</summary>

```Python
import vigra

atjson = """
{
   "axes":
   [
      {
         "key": "y",
         "typeFlags": 2
      }
   ]
}
"""
print(vigra.AxisTags.fromJSON(atjson)["y"].resolution)

</details>

So yes, the current code only adds the axis info required for ilastik to interpret it correctly.

k-dominik:
thx @tischi!!!

PR #142, opened 2026-02-10
Fix deploy with update Fiji executable and download

k-dominik:


k-dominik:
successfully deployed with this branch.

